PR #1, opened 2023-01-12
chore: add code to deploy protocol upgrades

ninocomputer:
### Description

This change will add the scripts we use to deploy new protocol contracts and to create governance proposals for upgrading the protocol.

### Other changes

n/a

### Tested
- Ran each deployment script on a Baklava successfully
- Ran governance proposal script on Baklava successfuly

### Related issues

- Fixes [97](https://github.com/mento-protocol/mento-core/issues/97)

### Backwards compatibility

n/a

### Documentation
- Added detailed [guide](https://www.notion.so/Core-Contracts-Deployment-Guide-3f68a4075e7f4a429f304ba8225d3c44) to wiki
- Added steps to readme


package.json
```diff
@@ -0,0 +1,60 @@
+{
```
ninocomputer, line 1: once the code is merged we should delete the existing package and release it again containing the readme

broadcast/00-CircuitBreaker.sol/62320/run-1673898399.json
```diff
@@ -0,0 +1,187 @@
+{
```
bowd, line 1: Are these the output of runs you've done when testing the branch? 
If so I would say we can remove them from this PR and we'll only keep the ones we actually care about as an outcome of https://github.com/mento-protocol/mento-core/issues/98

package.json
bowd: This is something where I haven't had time to think deeply about how best to structure things. If we deploy like this the README is actually the README of the `tooling` repo which is not what we want.

An option I have in mind is to have a subfolder, ' mento-core-ts' with the `package.json` and the deployment scripting and everything that has to do with the NPM package, and not have that at the root of the repo. But I can't tell how ugly that is.

Otherwise, we can consider actually not going down the `tooling` route as we initially thought, but have two separate repositories: 

- `mento-protocol/mento-deployment`
- `mento-protoco/mento-core-ts` 

Which both depend on `mento-protocol/mento-core`, and potentially `mento-protocol/mento-core-ts` might depend on `mento-protocol/mento-deployment` to access addresses (when we move away from the registry pattern).

I'm starting to think this might be a bit cleaner. Otherwise we're kinda doing a half monorepo, half not approach.

Thoughts @mento-protocol/core

src/IBreaker.ts
```diff
@@ -0,0 +1,201 @@
+/* Autogenerated file. Do not edit manually. */
```
bowd, line 1: Unsure if it would actually make sense to commit these, but it would probably make life easier for consumers.

bowd (commented):
Kinda rethinking this structure, lmk what y'all think.

broadcast/00-CircuitBreaker.sol/62320/run-1673898399.json
bayological: These runs are actually the ones we care about. I've also created a proposal for them on Baklava

package.json
bayological: @bowd @ninocomputer Resolving this now. This comment is no longer applicable to this repo. We have [a new issue](https://github.com/mento-protocol/mento-core-ts/issues/1) to create the `mento-core-ts` repo and can add the correct readme at the root of that repo.

src/IBreaker.ts
bayological: No longer relevant to this repo but will be added in the `mento-core-ts` repo

bowd (approved):

script/utils/Chain.sol
```diff
@@ -0,0 +1,83 @@
...
+  string public constant NETWORK_CELO_RPC = "celo_mainnet";
+  string public constant NETWORK_CELO_PK_ENV_VAR = "CELO_MAINNET_DEPLOYER_PK";
+
+  uint256 public constant NETWORK_BAKLAVA_CHAINID = 62320;
+  string public constant NETWORK_BAKLAVA_CHAINID_STRING = "62320";
+  string public constant NETWORK_BAKLAVA_RPC = "baklava";
+  string public constant NETWORK_BAKLAVA_PK_ENV_VAR = "BAKLAVA_DEPLOYER_PK";
+
+  uint256 public constant NETWORK_ALFAJORES_CHAINID = 44787;
+  string public constant NETWORK_ALFAJORES_CHAINID_STRING = "62320";
```
nvtaveras, line 25:
```suggestion
  string public constant NETWORK_ALFAJORES_CHAINID_STRING = "44787";
```

nvtaveras (commented):
LGTM, two small suggestions I caught and an optional nitpick

script/upgrades/00-MU01-mcmint/deploy/00-CircuitBreaker.sol
```diff
@@ -0,0 +1,74 @@
...
+
+    address[] memory __rateFeedIDs = new address[](0);
+    uint256[] memory __rateChangeThresholds = new uint256[](0);
+    uint256[] memory __cooldowns = new uint256[](0);
+
+    vm.startBroadcast(Chain.deployerPrivateKey());
+    {
+      medianDeltaBreaker = new MedianDeltaBreaker(
+        0,
+        0,
```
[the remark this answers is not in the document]
bayological: Good suggestion

nvtaveras (approved):
🚢

philbow61 (commented):
LGTM 🚢 exited to use this tomorrow

PR #9, opened 2023-02-08
feat: add mock usdc for testing

philbow61:
### Description

- created a mock USDC ERC20 contract 
- https://explorer.celo.org/baklava/address/0x5CB8FB4cF7247de5fD514243d72B5e90C1b7BAb7/read-contract#address-tabs
- https://explorer.celo.org/alfajores/address/0x8AF2F7958745Bdd40165F9F5D4492fc0dCd7D018/read-contract#address-tabs
- contracts are owned by the deployer address 0xF648678710d28025AE1A91d3fb08Ccc86dBFA493
- mint function is onlyOwner 

### Other changes

- N/A

### Tested

- tested the minting function

### Related issues

- Fixes https://github.com/mento-protocol/mento-core/issues/138

### Backwards compatibility

- N/A


ninocomputer (approved):
lgtm 👍🏼

contracts/MockUSDC.sol
```diff
@@ -0,0 +1,18 @@
+pragma solidity ^0.5.13;
+
+import { ERC20 } from "lib/mento-core/lib/openzeppelin-contracts/contracts/token/ERC20/ERC20.sol";
+import { ERC20Detailed } from "lib/mento-core/lib/openzeppelin-contracts/contracts/token/ERC20/ERC20Detailed.sol";
+import { Ownable } from "lib/mento-core/lib/openzeppelin-contracts/contracts/ownership/Ownable.sol";
+
+
+contract MockUSDC is ERC20, ERC20Detailed, Ownable {
+
+  constructor() public ERC20Detailed("mockUSDC", "USDC", 18) {
```
bayological, line 10: This is great, but I think we should do this slightly differently. Let's add a constructor that takes the name, symbol and decimals as parameters. Then rename this to MockERC20. It's likely that we'll have a need for additional ERC20 tokens we control for testing so this would make this a bit more generic.

bayological (changes requested):

contracts/MockUSDC.sol
philbow61: good point 👍 should be fixed now

script/upgrades/MU01-mcmint/deploy/MU01-04-Create-MockERC20.sol
```diff
@@ -0,0 +1,23 @@
+pragma solidity ^0.5.13;
+
+import { Script } from "script/utils/Script.sol";
+import { Chain } from "script/utils/Chain.sol";
+import { console2 } from "forge-std/Script.sol";
+
+import { MockERC20 } from "contracts/MockERC20.sol";
+
+contract MU01_CreateMockERC20 is Script {
```
bayological, line 9: Do you think this name would be better as CreateMockUSDC ?

bayological (changes requested):

bayological (approved):
LGTM 🛳️

PR #11, opened 2023-02-09
feat: add cUSD/USDcet exchange

### LINKED ISSUE #4: Create a USDC/cUSD pool in Baklava
**Description**

After deploying the dummy USDC token, we need to create a USDC/cUSD pool in Mento. The exchange should use the USDC/USD Oracle rates as a reference price (see the identifier for the pair in the [Oracle runbook](https://www.notion.so/Mento-Oracles-runbook-445b3fa6897544cf9eb53e859d7b7df9#a013838bab0747fda469f3d116cfd31e)) 

**Acceptance Criteria**
- [x] Governance proposal script is updated to include the USDC/cUSD pool
- [x] The governance proposal is simulated to ensure that adding the pool actually works

nvtaveras:
### Description

This adds a cUSD/USDcet exchange to the deployment scripts together with a USDCet -> cUSD swap test in the CGP simulation script. Also since the sortedOracles identifier for the USDC/USD rates is needed when configuring the exchange we added that to the dependencies.json file. 

### Other changes

We removed the `USDcet` address from the dependency.json since we couldn't find out the purpose of it. We assume it was a test contract deployed at some point which should now be the MockUSDcet contract.

### Tested

We couldn't run the simulation directly because the baklava contracts are already initialized, so what we did was re-run the MU01-mcmint deploy scripts on Baklava (createProxies, createNonUpgradeableContracts, createImplementations and governanceOwnershipTransfer) and then we ran a simulation using those newly deployed and not-yet initialized contracts to ensure that the exchange was created and that the swap was successful.

### Related issues

- Fixes #4

### Backwards compatibility

-

### Documentation

-


script/upgrades/dependencies.json
```diff
@@ -1,10 +1,10 @@
 {
   "42220": {},
   "62320": {
-    "USDCet": "0xfD46b16f86449b3e03e516A3632e1919A5CA1A53",
```
nvtaveras, line 4: @bowd is this useful for anything or was it only a mock token that was deployed at some point? we couldn't find any use for it so we removed it

ninocomputer (approved):
left some minor comments otherwise looks good!

script/test/Swap.sol
```diff
@@ -62,12 +77,39 @@ contract SwapTest is Script {
...
 
-    console2.log("Expected amount out:", amountOut);
+    console2.log("CELO -> cUSD swap Expected amount out:", amountOut);
 
     IERC20Metadata(contracts.celoRegistry("GoldToken")).approve(address(broker), 1e18);
     broker.swapIn(address(bpm), exchangeID, tokenIn, tokenOut, 1e18, amountOut - 1e17);
   }
 
+  function swapUSDcetForcUSD() public {
+    address trader = address(this);
```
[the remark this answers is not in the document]
philbow61: it can be any random address in this case we used the address of the contract

script/test/Swap.sol
```diff
@@ -7,18 +7,31 @@ import { Script } from "script/utils/Script.sol";
...
+
+  function balanceOf(address) external view returns (uint256);
+
+  function mint(address, uint256) external returns (bool);
+
+  function owner() external returns (address);
+}
+
 contract SwapTest is Script {
   IBroker broker;
```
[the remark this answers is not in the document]
philbow61: we fixed the linting issues that were related to the visibility of variables in the swap contract. The others should be fixed separately 👍

script/test/Swap.sol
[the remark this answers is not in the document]
ninocomputer: I'd still use another random address, using the contract's address this way is a bit confusing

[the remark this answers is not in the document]
tobikuhlmann: +1, or comment that it can be any random address

tobikuhlmann (changes requested):
looks good to me besides the `address(this)` as random address, see conversation

script/test/Swap.sol
```diff
@@ -7,18 +7,31 @@ import { Script } from "script/utils/Script.sol";
...
 
 import { IBroker } from "mento-core/contracts/interfaces/IBroker.sol";
+import { IStableToken } from "mento-core/contracts/interfaces/IStableToken.sol";
 import { IExchangeProvider } from "mento-core/contracts/interfaces/IExchangeProvider.sol";
 import { IERC20Metadata } from "mento-core/contracts/common/interfaces/IERC20Metadata.sol";
 import { BiPoolManager } from "mento-core/contracts/BiPoolManager.sol";
 import { IBiPoolManager } from "mento-core/contracts/interfaces/IBiPoolManager.sol";
 
+interface IMockERC20 {
+  // need to define this because we don't have a single interface with all this methods
```
bayological, line 17: Have you considered using the contract you've created? 
You can cast an address as a contract too, not just as an interface. The MockERC20 contract inherits from Ownable, which means it has the owner function, and it also inherits from ERC20, so it has all the other necessary functions.

bayological (changes requested):

script/upgrades/dependencies.json
```diff
@@ -1,10 +1,10 @@
...
-    "MockUSDCet": "0x5CB8FB4cF7247de5fD514243d72B5e90C1b7BAb7"
+    "MockUSDCet": "0x5CB8FB4cF7247de5fD514243d72B5e90C1b7BAb7",
+    "USDCUSDRateFeedAddr": "0xA1A8003936862E7a15092A91898D69fa8bCE290c"
```
bayological, line 5: How was this address generated?

script/test/Swap.sol
```diff
@@ -29,6 +42,7 @@ contract SwapTest is Script {
     // Get proxy addresses of the deployed tokens
     cUSD = contracts.celoRegistry("StableToken");
     cEUR = contracts.celoRegistry("StableTokenEUR");
+    usdCet = contracts.dependency("MockUSDCet");
```
bayological, line 45: Is there a reason we're pulling this from the dependency instead of the [deployed contracts](https://github.com/mento-protocol/mento-deployment/blob/971caff93f794079539bdb84c5134fb79c07e9eb/script/utils/Contracts.sol#L74)?

script/upgrades/dependencies.json
```diff
@@ -1,10 +1,10 @@
 {
```
bayological, line 1: It looks like the intention of this file is to store references to any dependencies that are not in our control. Contracts we are deploying as part of our updates should not be here. The addresses of any contracts we're deploying will be in the broadcast json files. These are loaded into the context of the script by first calling the helper function [contracts.load()](https://github.com/mento-protocol/mento-deployment/blob/971caff93f794079539bdb84c5134fb79c07e9eb/script/utils/Contracts.sol#L30) which takes the name of the deployment script and its timestamp as arguments. 

You can then pull the address for any contract that's been deployed as part of the loaded deployment scripts by calling [contracts.deployed()](https://github.com/mento-protocol/mento-deployment/blob/971caff93f794079539bdb84c5134fb79c07e9eb/script/utils/Contracts.sol#L74). This simplifies things as we always get the correct address and we do not have to keep updating this file. 

This may not be clear so it would be good to add this to the readme of the repo

script/upgrades/dependencies.json
tobikuhlmann: not sure if it's useful but the baklava deployment script fails if this is deleted, can you put it back in?

nvtaveras: How did you ran it? It doesn't fail for us without it

script/test/Swap.sol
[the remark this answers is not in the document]
nvtaveras: updated

script/test/Swap.sol
nvtaveras: ah yeah didn't think of this  🤦🏽‍♂️ updated

script/upgrades/dependencies.json
nvtaveras: This is the identifier address for `USDCUSD` that is currently used for the sortedOracles. Since we already passed a proposal with this identifier, calculating this in the code with a function seems dangerous, as that could have a bug and end up using a different address from what we have on-chain.

script/upgrades/dependencies.json
nvtaveras: We just made a change to rename `mockUSDCet` to just `USDcet`. This is such that the deployment scripts can work on all networks. For Baklava and Alfajores, the address will be our deployed MockUSDcet contract, while in mainnet it will be the actual USDCet contract address from wormhole (therefore we need to have the addresses hardcoded in the dependencies.json file)

script/test/Swap.sol
nvtaveras: replied in your other comment

script/upgrades/dependencies.json
tobikuhlmann: yup the script is loading `USDCet` so now it works fine

tobikuhlmann (approved):

bayological (approved):

PR #19, opened 2023-02-23
feat: add script to get secrets

### LINKED ISSUE #8: Google Cloud private key management
**Description**

We need to standardize how we access private keys in our development/deployment environment.
We currently have:

- Mento Deployer private key, which can be the same across all environments
- Baklava Approver private key
- Baklava Voter private key
- Alfajores Approver private key
- Alfajores Voter private key

These are required by the Foundry scripts and the [testnet CGP helper script.](https://github.com/mento-protocol/mento-deployment/blob/main/script/approve_and_execute.sh) and are currently loaded in a non-standard way across the board.

**Acceptance criteria**

The outcome of this issue would be to have a standard way of pulling the secrets from Google Cloud Secret Manager and having them in a format that can be consumed by all the scripts. This can be achieved by having a script that uses the `gcloud` command to query the secrets from google cloud and, if not already present, adds them to the end of the `.env` file which is ignored by git and would end up looking like this:
```
CELO_RPC_URL='https://forno.celo.org'
BAKLAVA_RPC_URL='https://baklava-forno.celo-testnet.org'
MENTO_DEPLOYER_PK=...
BAKLAVA_APPROVER_PK=...
BAKLAVA_VOTER_PK=...
...
```

Forge out loads the `.env` file as far as I understand so that would mean that these are now accessible to Foundry scripts, and we should change the [testnet CGP helper script.](https://github.com/mento-protocol/mento-deployment/blob/main/script/approve_and_execute.sh) to start with a `source .env` to make use of the keys.

Brownie points: modify the `approve_and_execute.sh` script to take two arguments the environment to target and the proposal ID, and then use the correct keys depending on that.

Readme should be updated to explain the usage.

bayological:
### Description

- This change adds a script that pulls specified secrets from GCP and adds them to a local .env file.

### Other changes

 - N/A

### Tested

- Ran script to verify it works as expected

### Related issues

- Closes #8 

### Backwards compatibility

 - N/A

### Documentation

- Added line to readme to show useage


tobikuhlmann (approved):
looks good to me

philbow61 (approved):
Nice 🤩

PR #34, opened 2023-03-29
feat(nonupgradable contracts): feature/addDefaultBreakerValues

tobikuhlmann:
### Description

This PR adds real default values to median and value breaker deployment.

To verify the default values and how I chose them, see the issue description: https://github.com/mento-protocol/mento-core/issues/174

### Other changes

-

### Tested
- Yes, test-deployed to Baklava 

### Related issues

- Fixes https://github.com/mento-protocol/mento-core/issues/174

### Backwards compatibility

yes should be fully

### Documentation
- https://github.com/mento-protocol/mento-core/issues/174

### DIFF
--- a/script/upgrades/MU01/deploy/MU01-01-Create-Nonupgradeable-Contracts.sol
+++ b/script/upgrades/MU01/deploy/MU01-01-Create-Nonupgradeable-Contracts.sol
@@ -5,6 +5,7 @@ import { console2 } from "forge-std/Script.sol";
 import { Script } from "script/utils/Script.sol";
 import { Chain } from "script/utils/Chain.sol";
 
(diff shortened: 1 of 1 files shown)


sissnad:
My understanding was that setting `CoolDown` to 0 means [manual reset ](https://github.com/mento-protocol/mento-core/blob/f430e91d2a3783242d10a9098c1bb4ed8f4e36f5/contracts/BreakerBox.sol#L330)is required which is the safest and most restrictive setting. I believe this is why it was set to 0 initially

sissnad:
Similar argument for setting the threshold to 0

tobikuhlmann:
closing as agreed upon 0 default values

PR #51, opened 2023-05-02
chore: add new bridgedUSDC to reserve

### LINKED ISSUE #42: Create testnet-only governance proposal script to update MockUSDC
**Description**

Create a testnet-only governance proposal script that updates all dependencies to use the new MockUSDC contract with 6 decimals that will be deployed as part of https://github.com/mento-protocol/mento-deployment/issues/41. See the full plan [here](https://www.notion.so/cUSD-to-USDC-swaps-43904bb61f7e496fb827b881668d53d2).

**Acceptance Criteria**

- [ ] Testnet-only governance proposal script is created
- [ ] Script performs required changes:
  - [ ] Remove old MockUSDC as a reserve currency
  - [ ] Add new MockUSDC as a reserve currency
  - [ ] Destroy the cUSD-USDC pool
  - [ ] Create the cUSD-USDC pool (using the new MockUSDC address)
- [ ] Script is tested and verified on testnets

ninocomputer:
### Description

Created a new testnet only script which performs the following:
- delete old cUSD/USDC pool
- add the pool with updated BridgedUSDC address
- add new bridgedUSDC with 6 decimals as a reserve collateral
- configure and add spending ratio
- remove the old one with 18 decimals

The script was executed on Baklava - [transaction details](https://explorer.celo.org/baklava/tx/0x50b39c3e22cd4511c720a09b76c4f14ec1d66dc0d9b1a4ea32efe217e5ca305d/internal-transactions?block_number=17018323&index=50&items_count=50&transaction_index=0)

### Other changes

n/a

### Tested

n/a

### Related issues

- Fixes #42 

### Backwards compatibility

n/a

### Documentation

n/a


bowd (changes requested):
@ninocomputer I thin we misunderstood each other when we chatted about this. Reusing the Phase2 script for this doesn't make sense because it is intended to be used on mainnet as well. While I understand that it will be a noop on mainnet, I think it's messy to combine the two. What I meant when we chatted is that we should use the phase2 config when recreating the cUSD/USDC pool in this testnet-only script that would only do the bridged USDC swap and recreate. 

I'd suggest extracting the logic in a new script.

bin/cgp.sh
```diff
@@ -8,7 +8,7 @@
 #               -p <phase_number>               -- phase number of the upgrade (1)
 #               -s                              -- simulate the proposal (optional)
 #               -f                              -- use forked network (optional)
-# Example: ./bin/cgp.sh -n baklava -u MU01 -p 1 
+# Example: ./bin/cgp.sh -n baklava -u MU01 -p 1 || yarn cgp -n baklava -u MU01 -p 1_testnet_patch
```
bayological, line 11: I have a couple of thoughts on this. 
-  @ninocomputer With this naming, the simulation script will not work as it expects the phase to be a uint8
- @bowd I feel this could get messy in future especially if we were to require multiple patches in one phase. e.g. another issue is discovered after the deployment of a patch. Should we switch to something like SemVer going forward? So CGP scripts for a future MU could look like this -> MU02-CGP-1-0-1

bowd: Maybe, tho just pure semver might not be enough. For example this script is only meant to be run on testnets so we would need to differentiate somehow, or have some minor versions that do nothing in production environments or something weird like that. 

Initially I thought of combining this with the other `Phase1-patch1` script and have a conditional in the `prepare()` to add some transactions to the CGP only on testnets, but I thought that might be ugly and mixing up some testnet-only logic with production logic, so decided having two separate ones made sense. but yeah maybe we can do something like `semver + network modifier`. 

Let's revisit tho as we learn more about what we actually need to do in these scenarios, for now I think it's fine to keep it simple.

bowd (approved):

philbow61 (approved):
LGTM 👍

Issue #59, opened 2023-06-06
Create deployment script for new MedianDeltaBreaker


bayological:
**Description**

Given the recent updates to the MedianDeltaBreaker, we require a refined deployment script that we can ren across all environments to deploy the updated version.

**Acceptance Criteria**

- New deployment script created and added to the deploy folder
- The script should deploy the breaker, using default configuration values, then transfer ownership to governance.

bayological:
**Story Points: 2**
  [See the discussion](https://action.parabol.co/meet/nJxBsGtXS8/estimate/5) in Sprint Poker #​​12

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

PR #68, opened 2023-06-14
chore: add BreakerBox deploy script

ninocomputer:
### Description

write a deploy script for BreakerBox

### Other changes

- fixed test in DeploymenyChecks that wouldn't compile 
- update the mento-core library
- commented out incompatible logic and kept it there for context  
- lint

### Tested

n/a

### Related issues

- Fixes #60 

### Backwards compatibility

Not backwards compatible

### Documentation

n/a

PR #69, opened 2023-06-16
feat: add medianDeltaBreakear deployment script

nvtaveras:
### Description

This adds the script for deploying the median delta breaker with default/empty configuration and transferring it to governance so that we can later re-configure it with the right parameters in the MU02 activation CGP.

### Other changes

-

### Tested

Ran the deployment simulation locally

### Related issues

- Fixes https://github.com/mento-protocol/mento-deployment/issues/59

### Backwards compatibility

Yes, since the old constructor didn't change

### Documentation

-

### DIFF
--- a/script/upgrades/MU02/MU02-01-Deploy-MedianDeltaBreaker.sol
+++ b/script/upgrades/MU02/MU02-01-Deploy-MedianDeltaBreaker.sol
@@ -1,0 +1,49 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.5.13;
+
+import { console2 } from "forge-std/Script.sol";
+import { Script } from "script/utils/Script.sol";
+import { Chain } from "script/utils/Chain.sol";
+
+import { MedianDeltaBreaker } from "mento-core/contracts/MedianDeltaBreaker.sol";
+import { ISortedOracles } from "mento-core/contracts/interfaces/ISortedOracles.sol";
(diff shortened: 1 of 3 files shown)


ninocomputer (approved):

script/upgrades/MU02/MU02-01-Deploy-MedianDeltaBreaker.sol
bayological, line 1: The deployment script is fine, however the location and file name are not in line with the convention. The folder under upgrades is intended to collect all deployment scripts related to a particular upgrade. We only have one upgrade, MU01, and this folder should contain everything related to MC Mento developments.

The location of this script should be ``` upgrades/MU01/deploy ``` and the name should be ``` MU01-05-Deploy-MedianDeltaBreaker.sol ``` (Number should be the next one up from max)

bayological (changes requested):

script/upgrades/MU02/MU02-01-Deploy-MedianDeltaBreaker.sol
nvtaveras: Chatted with Bayo offline about this, decided that we will stick to MU01 for all the new deployment scripts. I have also moved the breakerBox deployment script to the MU01 folder (https://github.com/mento-protocol/mento-deployment/pull/68)

bayological (approved):

PR #92, opened 2023-07-10
chore: refactor configs to improve readability

### LINKED ISSUE #94: Restructure MU Config for readability
**Description**

Our configuration has grown in complexity and has become less readable and harder to verify, it also enforces a 1:1 link between pools and rateFeeds, both being configured in the same struct. This will not be the case in the future so spending some time now to refactor and cleanup feels warranted.

**Acceptance criteria**

- [x] Configs are rewritten and broken up into multiple structs for readability
- [x] Everything is retrofitted and verified

bowd:
### Description

Refactor configs to improve their readability:
- Break up the `PoolConfiguration` into `PoolConfig` (contains BiPoolManager config and trading limits) and `RateFeedConfig` containing circuit breaker configs for a rate feed, this will come in especially handy when we start adding rate feeds that aren't followed by a pool but are required as dependant rate feeds (like EUROC/EUR). 
- Add a `TradingLimitsConfig` to further DRYup the `PoolConfig`

### Other changes

N/A

### Tested

Did some before and after simulation traces.

### Related issues

 - Fixes #94  

### Backwards compatibility

No, but previous MUs have been updated.

### Documentation

N/A


script/upgrades/MU03/MU03Checks.sol
```diff
@@ -362,31 +360,27 @@ contract MU03Checks is Script, Test {
     console2.log("\tBreakers set with trading mode 3");
 
     // verify that rate feed dependencies were configured correctly
-    address cEurDependency = BreakerBox(breakerBox).rateFeedDependencies(cEURUSDCConfig.referenceRateFeedID, 0);
-    address cBrlDependency = BreakerBox(breakerBox).rateFeedDependencies(cBRLUSDCConfig.referenceRateFeedID, 0);
-    require(cEurDependency == cUSDUSDCConfig.referenceRateFeedID, "cEUR/USDC dependency not set correctly");
-    require(cBrlDependency == cUSDUSDCConfig.referenceRateFeedID, "cBRL/USDC dependency not set correctly");
+    address cEurDependency = BreakerBox(breakerBox).rateFeedDependencies(config.cEURUSDC.referenceRateFeedID, 0);
```
bowd, line 363: Because this is more RateFeed related and not pool related I would use those configs instead, i.e.:
```solidity
address USDCEURDependency0 = BreakerBox(breakerBox).rateFeedDependencies(config.USDCEUR.rateFeedID, 0);
require(USDCEURDependency0, config.USDCUSD.rateFeedID);
```

Also naming it cEUR dependency I find potentially confusing. What the system is actually saying is that the USDCEUR rate feed is dependant on the USDCUSD here, the pools themselves, and thus cEUR doesn't really factor in necessarily.

script/upgrades/MU03/MU03Checks.sol
```diff
@@ -401,110 +395,82 @@ contract MU03Checks is Script, Test {
...
 
         // verify smoothing factor
-        if (smoothingFactor != poolConfigs[i].smoothingFactor) {
-          console2.log(
-            "MedianDeltaBreaker smoothing factor not set correctly for rate feed %s",
-            poolConfigs[i].referenceRateFeedID
-          );
-          revert("MedianDeltaBreaker smoothing factor not set correctly for all rate feeds");
+        if (smoothingFactor != rateFeed.medianDeltaBreaker0.smoothingFactor) {
+          console2.log("MedianDeltaBreaker smoothing factor not set correctly for cUSD/USDC %s", rateFeed.rateFeedID);
```
bowd, line 421: Same here, we're not setting the breakers on `cUSD/USDC`. We're setting the breakers on `USDC/USD` oracles, that value is then used to power the `cUSD/USDC` pools, but the breakers don't act on or care about the `cUSD/USDC` pool's existence or rate.

script/upgrades/MU03/MU03Checks.sol
```diff
@@ -744,8 +710,8 @@ contract MU03Checks is Script, Test {
   ) internal view {
     if (currentThreshold != expectedThreshold) {
       if (isValueDeltaBreaker) {
-        console2.log("ValueDeltaBreaker rate change threshold not set correctly for rate feed %s", rateFeedID);
-        revert("ValueDeltaBreaker rate change threshold not set correctly for all rate feeds");
+        console2.log("ValueDeltaBreaker rate change threshold not set correctly for cUSD/USDC %s", rateFeedID);
+        revert("ValueDeltaBreaker rate change threshold not set correctly for all cUSD/USDC");
```
bowd, line 714: ditto, and I think here we probably don't want `cUSD/USDC` written at all right? this is a function called with a rateFeedID that can be any one.

ninocomputer: since we use valueDeltaBreaker only for USDC/USD rate feed I decided to make it specific in mu03. should I change it?

philbow61 (approved):
🚀 💪

script/upgrades/MU01/MU01.sol
```diff
@@ -222,39 +204,46 @@ contract MU01 is IMentoUpgrade, GovernanceScript {
...
-    if (partialReserveProxy._getImplementation() == address(0)) {
-      Reserve reserve = Reserve(contracts.deployed("Reserve"));
-      transactions.push(
-        ICeloGovernance.Transaction(
-          0,
-          address(partialReserveProxy),
-          abi.encodeWithSelector(
-            partialReserveProxy._setAndInitializeImplementation.selector,
-            reserve,
+    {
```
nvtaveras, line 207: did you had to do this because of a stack to deep error or is there another reason?

script/upgrades/MU03/Config.sol
```diff
@@ -13,198 +13,274 @@ library MU03Config {
...
+    });
+  }
+
+  function cBRLCelo_PoolConfig(Contracts.Cache storage contracts) internal view returns (Config.Pool memory config) {
+    config = Config.Pool({
       asset0: contracts.celoRegistry("StableTokenBRL"),
       asset1: contracts.celoRegistry("GoldToken"),
       isConstantSum: false,
-      spread: FixidityLib.newFixedFraction(25, 10000), // 0.0025
+      spread: FixidityLib.newFixedFraction(25, 10_000), // 0.0025
```
nvtaveras, line 136: do you also want to change this in the other pool configs to be consistent?

nvtaveras (approved):
LGTM 🚢

script/upgrades/MU01/MU01.sol
bowd: Yeah stack too deep :(

PR #97, opened 2023-07-19
chore: upgrade sortedoracles

### LINKED ISSUE #96: Update MU03 deployment scripts and CGP to upgrade the SortedOracles implementation
**Acceptance Criteria**

- [ ] write the deploy script 
- [ ] update the SortedOracles implementation in the gov script

ninocomputer:
### Description

- [ ] Updated deployment script to include SortedOracles
- [ ] Extended MU03checks 
- [ ] Deployed on alfajores/baklava

### Other changes

n/a

### Tested

governance simulation works

### Related issues

- Fixes #96

### Backwards compatibility

n/a

### Documentation

n/a


script/upgrades/MU03/MU03Checks.sol
```diff
@@ -17,6 +17,7 @@ import { IBroker } from "mento-core-2.2.0/interfaces/IBroker.sol";
 import { IERC20Metadata } from "mento-core-2.2.0/common/interfaces/IERC20Metadata.sol";
 
 import { BiPoolManagerProxy } from "mento-core-2.2.0/proxies/BiPoolManagerProxy.sol";
+import { SortedOraclesProxy } from "mento-core-2.0.0/proxies/SortedOraclesProxy.sol";
```
bowd, line 20: Let's use the `2.2.0` consistently in `MU03`, even tho in practice this ends up being the same contract in this case, if we mix up different versions we might introduce subtle bugs.

script/upgrades/MU03/MU03Checks.sol
```diff
@@ -134,6 +142,21 @@ contract MU03Checks is Script, Test {
     console2.log("\tBiPoolManagerProxy has a correct implementation address 🫡");
   }
 
+  function verifySortedOracles() internal view {
+    SortedOraclesProxy soProxy = SortedOraclesProxy(sortedOraclesProxy);
```
bowd, line 146: This is correct but I kinda want us to move away from using named proxies so for the sake of that I'd import the base `Proxy` contract instead and cast it as that here, just to future-proof the system.

bowd (changes requested):
Small nits ⛩️

script/upgrades/MU03/MU03Checks.sol
ninocomputer: good catch! it was the co-pilot 😤

bowd (approved):

philbow61 (approved):

PR #104, opened 2023-08-21
Fix/mu03 exchange destruction

philbow61:
### Description
- this Pr removes the destruction and recreation of all pools and only destroys the cUSD/USDC pool since that one actually needs to be destroyed and recreated. 
- also it changes the way constant product pair swaps are tested. The Calculation from now on uses the current median rate instead of the buckets as these can be largely outdated and cause failing tests. 

### Other changes

- reordered the implementation updates to represent the call order

### Tested

- simulation is green 

### Related issues

- Fixes #[issue number here]


nvtaveras (approved):
🚢

ninocomputer (approved):
lgtm

PR #106, opened 2023-08-24
fix: remove unnecessary Value Delta Breaker transactions

### LINKED ISSUE #109: Remove unneeded ValueDeltaBreaker transactions in MU03
We don't need to reconfigure the ValueDeltaBreaker again for USDC/USD in MU03 since the breaker&config hasn't changed

philbow61:
### Description
- Since the ValueDeltaBreaker hasn't changed in MU03 and the breaking logic is already set for USDC/USD there is no need for setting it again. We only need to configure the breaker for the new EUROC/EUR rate.

### Tested

- The values on mainnet for the USDC/USD rate feed are equal to the once in the config. 
- I ran an isolated simulation on baklava with only the valueDeltaBreaker part + the checks  
- also we are going to run the whole script on alfajores where the state is more simular to mainnet 

### Related issues

- Fixes #109 

### Backwards compatibility

- the configuration already is configured with the same values

### DIFF
--- a/script/upgrades/MU03/MU03.sol
+++ b/script/upgrades/MU03/MU03.sol
@@ -499,47 +499,41 @@ contract MU03 is IMentoUpgrade, GovernanceScript {
   }
 
   function proposal_configureValueDeltaBreaker0(MU03Config.MU03 memory config) public {
-    // Set reference value for USDC/USD, EUROC/EUR rate Feed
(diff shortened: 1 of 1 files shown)


ninocomputer (approved):

nvtaveras (approved):
👍

PR #107, opened 2023-08-25
feat: set spending ratio for EUROC

philbow61:
### Description

we forgot to set the `Daily Spending Ratio For Collateral Asset` for Euroc. This PR adds the needed transaction plus the check. I also  ran the script on baklava 
image

### Tested

- ran the simulation 
- ran the script + checks 
- verified the value afterward


nvtaveras (approved):
LGTM

ninocomputer (approved):

PR #115, opened 2023-08-29
feat: create eXOF governance script

### LINKED ISSUE #77: Create eXOF launch governance script

bayological:
### Description

This PR adds the transactions required for the eXOF governance proposal. 

### Other changes

- Added helper function in contracts cache to load dependent contracts without logging

### Tested

- Simulated proposal successfully on testnet
- Added checks to verify governance changes
- Added tests to verify swaps in new eXOF pools

### Related issues

- Closes #77 

### Backwards compatibility

- N/A

### Documentation

- N/A


script/upgrades/eXOF/eXOFChecks.swap.sol
```diff
@@ -0,0 +1,242 @@
...
+import { SortedOracles } from "mento-core-2.2.0/oracles/SortedOracles.sol";
+
+import { eXOFChecksBase } from "./eXOFChecks.base.sol";
+import { eXOFConfig, Config } from "./Config.sol";
+
+contract eXOFChecksSwap is eXOFChecksBase {
+  using FixidityLib for FixidityLib.Fraction;
+  using Contracts for Contracts.Cache;
+
+  struct DoSwapInArgs {
```
bayological, line 27: Remove this.

script/upgrades/eXOF/eXOFChecks.verify.sol
```diff
@@ -0,0 +1,504 @@
...
+          pool.config.stablePoolResetSize,
+          poolConfig.stablePoolResetSize
+        );
+        revert("stablePoolResetSize of pool does not match the expected stablePoolResetSize. See logs.");
+      }
+    }
+    console.log("🟢 Pool config is correctly configured 🤘🏼");
+  }
+
+  function verifyTradingLimits(eXOFConfig.eXOF memory config) internal view {
```
philbow61, line 260: I think we are just verifying trading limits on asset 0 here since we introduce an exchange with limits configured for both assets in this deployment I would add a check for asset1 trading limits as well

script/upgrades/eXOF/eXOFChecks.swap.sol
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    FixidityLib.Fraction memory maxTolerance = FixidityLib.newFixedFraction(25, 1000);
+    uint256 threshold = FixidityLib.newFixed(estimatedAmountOut).multiply(maxTolerance).fromFixed();
+
+    assertApproxEq(amountOut, estimatedAmountOut, threshold);
+    doSwapIn(exchangeID, trader, tokenIn, tokenOut, amountIn, amountOut);
+  }
+
+  function testAndPerformConstantSumSwap(
```
philbow61, line 173: In MU03 we refactored this function and simplified it I would suggest using that one. If not I would rename the USDC part to EUROC since we only perform Constant sum EUROC swaps here

philbow61 (changes requested):
LGTM good job 👍 
just some nits here and there

broadcast/eXOF-01-Create-Implementations.sol/44787/run-latest.json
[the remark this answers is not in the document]
bayological: This was @bowd, the only deployment I ran was for the non-upgradeable contracts on Baklava.

script/upgrades/eXOF/Config.sol
```diff
@@ -0,0 +1,190 @@
...
+  using Contracts for Contracts.Cache;
+
+  struct eXOF {
+    // Pools
+    Config.Pool eXOFCelo;
+    Config.Pool eXOFEUROC;
+    Config.Pool[] pools;
+    // Rate Feeds
+    Config.RateFeed CELOXOF;
+    Config.RateFeed EUROXOF;
```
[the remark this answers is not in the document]
bayological: 🤝

script/upgrades/eXOF/eXOF.sol
```diff
@@ -0,0 +1,601 @@
...
+pragma experimental ABIEncoderV2;
+
+import { GovernanceScript } from "script/utils/Script.sol";
+import { console2 as console } from "forge-std/Script.sol";
+import { Contracts } from "script/utils/Contracts.sol";
+import { Chain } from "script/utils/Chain.sol";
+import { Arrays } from "script/utils/Arrays.sol";
+
+import { FixidityLib } from "mento-core-2.2.0/common/FixidityLib.sol";
+import { IBiPoolManager } from "mento-core-2.2.0/interfaces/IBiPoolManager.sol";
```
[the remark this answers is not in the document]
bayological: 🙅‍♂️ Switching to only the contract adds nothing. Instead, I've removed imports of the concrete implementations as they were not being used. In the case of the BiPoolManager, we actually need to reference the interface, as that is where the structs(PoolConfig) are defined.

philbow61 (approved):
🚢

script/upgrades/eXOF/eXOFChecks.verify.sol
```diff
@@ -0,0 +1,517 @@
...
+    verifyPoolConfig(config);
+    verifyTradingLimits(config);
+  }
+
+  function verifyPoolExchange(eXOFConfig.eXOF memory config) internal view {
+    bytes32[] memory exchanges = BiPoolManager(biPoolManagerProxy).getExchangeIds();
+
+    // check configured pools against the config
+    require(
+      exchanges.length == config.pools.length + 7,
```
nvtaveras, line 141: maybe put the `7` into a variable to not have the magic constant?

script/upgrades/eXOF/eXOFChecks.verify.sol
```diff
@@ -0,0 +1,517 @@
...
+      Config.Pool memory poolConfig = config.pools[i];
+
+      // verify asset0 of the deployed pool against the config
+      if (pool.asset0 != poolConfig.asset0) {
+        console.log(
+          "The asset0 of deployed pool: %s does not match the expected asset0: %s.",
+          pool.asset0,
+          poolConfig.asset0
+        );
+        revert("asset0 of pool does not match the expected asset0. See logs.");
```
nvtaveras, line 158: maybe let's put statements like these into `require()` instead? like how it's done in a few lines below in L191, L196. This is inconsistent across other functions as well

script/upgrades/eXOF/eXOFChecks.verify.sol
```diff
@@ -0,0 +1,517 @@
...
+      }
+      // verify asset0 is always a stable asset
+      require(
+        pool.asset0 == cUSD || pool.asset0 == cEUR || pool.asset0 == cBRL || pool.asset0 == eXOF,
+        "asset0 is not a stable asset in the exchange"
+      );
+      // verify asset1 is always a collateral asset
+      require(
+        pool.asset1 == celoToken || pool.asset1 == bridgedUSDC || pool.asset1 == bridgedEUROC,
+        "asset1 is not CELO or bridgedUSDC in the exchange"
```
nvtaveras, line 198:
```suggestion
        "asset1 is not CELO or bridgedUSDC or bridgedEUROC in the exchange"
```

script/upgrades/eXOF/eXOFChecks.verify.sol
```diff
@@ -0,0 +1,517 @@
...
+            "The pricing module of deployed pool: %s does not match the expected pricing module: %s.",
+            address(pool.pricingModule),
+            constantProduct
+          );
+          revert("pricing module of pool does not match the expected pricing module. See logs.");
+        }
+      }
+      // verify asset0 is always a stable asset
+      require(
+        pool.asset0 == cUSD || pool.asset0 == cEUR || pool.asset0 == cBRL || pool.asset0 == eXOF,
```
nvtaveras, line 192: now that we have the `Arrays.contains` function we could also put all of these in an array and use that instead. Similar for the collateral assets.

script/upgrades/eXOF/eXOFChecks.verify.sol
```diff
@@ -0,0 +1,517 @@
...
+        "asset1 is not CELO or bridgedUSDC in the exchange"
+      );
+    }
+    console.log("🟢 PoolExchange correctly configured 🤘🏼");
+  }
+
+  function verifyPoolConfig(eXOFConfig.eXOF memory config) internal view {
+    for (uint256 i = 0; i < config.pools.length; i++) {
+      bytes32 exchangeId = getExchangeId(config.pools[i].asset0, config.pools[i].asset1, config.pools[i].isConstantSum);
+      IBiPoolManager.PoolExchange memory pool = BiPoolManager(biPoolManagerProxy).getPoolExchange(exchangeId);
```
nvtaveras, line 207: maybe something like `deployedPool` or `actualPool` would be a better name here, otherwise the comparison look a bit confusing (i.e. `pool.config.spread.unwrap() != poolConfig.spread.unwrap()`). Similarly in all the other functions that check the actual pools vs the pool configuration.

script/upgrades/eXOF/eXOFChecks.verify.sol
```diff
@@ -0,0 +1,517 @@
...
+
+        // verify rate change threshold
+        verifyRateChangeTheshold(
+          rateChangeThreshold,
+          rateFeed.valueDeltaBreaker0.threshold.unwrap(),
+          rateFeed.rateFeedID,
+          true
+        );
+
+        // verify refernece value
```
nvtaveras, line 436:
```suggestion
        // verify reference value
```

nvtaveras (changes requested):
Did a first pass through, some small things but the main ones are:

- Missing deployment checks for the constitution changes in the proposal
- eXOF shouldn't be added to the main reserve yet (I'll check this with others and confirm)

script/upgrades/eXOF/eXOF.sol
```diff
@@ -0,0 +1,599 @@
...
+
+  /**
+   * @notice This function creates the transactions to configure the trading limits.
+   */
+  function proposal_configureTradingLimits(eXOFConfig.eXOF memory config) private {
+    address brokerProxyAddress = contracts.deployed("BrokerProxy");
+    for (uint256 i = 0; i < config.pools.length; i++) {
+      Config.Pool memory poolConfig = config.pools[i];
+
+      if (Config.tradingLimitConfigToFlag(poolConfig.asset0limits) > 0) {
```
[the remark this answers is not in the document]
bayological: I agree, but I would not throw an error here; that is not consistent with the other proposal functions but also the whole point of the simulation checks is to do things like that. IMO, we can just remove these. But I like the intention behind this, possibly, in future, if we keep building on this deployment tooling, we could implement more generic functions that look a little like this for the functions we regularly call, such as creating exchanges & configuring trading limits.

script/upgrades/eXOF/eXOF.sol
```diff
@@ -0,0 +1,599 @@
...
+      ICeloGovernance.Transaction(
+        0,
+        medianDeltaBreaker,
+        abi.encodeWithSelector(
+          MedianDeltaBreaker(0).setSmoothingFactor.selector,
+          config.CELOXOF.rateFeedID,
+          config.CELOXOF.medianDeltaBreaker0.smoothingFactor
+        )
+      )
+    );
```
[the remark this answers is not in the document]
bayological: There's nothing in notion about the final values for the breakers, but also, @sissnad hasn't verified the configuration, so it's possible the value may not be zero. I will leave this unchanged until he has checked. If it turns out there's nothing to set, then I'll remove it.

script/upgrades/eXOF/eXOFChecks.swap.sol
```diff
@@ -0,0 +1,233 @@
...
+    bytes32 exchangeID,
+    address trader,
+    address tokenIn,
+    address tokenOut,
+    uint256 amountIn,
+    address rateFeedID,
+    bool isBridgedStable
```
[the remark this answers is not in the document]
bayological: We need this parameter to correctly calculate the estimated output amount of the swap whilst taking into account the different decimal precisions of the tokens. But you are right in that the name may not make sense. Renamed it for clarity. :)

script/upgrades/eXOF/eXOFChecks.swap.sol
```diff
@@ -0,0 +1,233 @@
...
+import { eXOFChecksBase } from "./eXOFChecks.base.sol";
+import { eXOFConfig, Config } from "./Config.sol";
+
+contract eXOFChecksSwap is eXOFChecksBase {
+  using FixidityLib for FixidityLib.Fraction;
+  using Contracts for Contracts.Cache;
+
+  constructor() public {
+    new PrecompileHandler();
+    setUp();
```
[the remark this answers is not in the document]
bayological: Running setup in the eXOFChecks was left over from the refactor. And that inheritance would not work. The proposal simulation relies on the checks for an upgrade to have the same name as the upgrade with 'checks' appended. So 'eXOFChecksVerify' would not be picked up by the simulation helper function.

script/upgrades/eXOF/eXOFChecks.verify.sol
bayological: 🙅 Consistency isn't the point here. The log statements are here to add additional information using string interpolation to help debug any failed checks. Solidity doesn't support string interpolation in require statements, so it's best this remains this way.

script/upgrades/eXOF/eXOFChecks.verify.sol
bayological: 🙅 We could, and the function would make the code more concise. However, switching to it doesn't add much and would make the intention of the require statement less clear, as we would need to introduce an additional data structure to hold the stables. The current implementation with multiple OR conditions is relatively clear and directly shows the intention of the code.

script/upgrades/eXOF/eXOFChecks.verify.sol
bayological: The naming choice is definitely poor, however reading the code where the variables are populated would clarify any confusion.

script/upgrades/eXOF/eXOF.sol
```diff
@@ -0,0 +1,599 @@
...
+   * @notice configure eXOF constitution parameters
+   * @dev see cBRl GCP(https://celo.stake.id/#/proposal/49) for reference
+   */
+  function proposal_configureEXOFConstitutionParameters() private {
+    bytes4[] memory functionSelectors = Arrays.bytes4s(
+      getSelector("setRegistry(address)"),
+      getSelector("setInflationParameters(uint256,uint256)"),
+      getSelector("transfer(address,uint256)"),
+      getSelector("transferWithComment(address,uint256,string)"),
+      getSelector("approve(address,uint256)")
```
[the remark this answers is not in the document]
bayological: In your example, you're missing the complete signature, you've only used the function name and skipped the arg types, which is not the same as the code you commented on. 

Using chisel with the correct signatures here's what I got:
image

And this is the CGP:
image

Good catch on the checks, they have been added.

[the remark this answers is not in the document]
nvtaveras: Thanks, I could have sworn I copied the parameters into chisel as well 🤦🏽‍♂️ the checks + constitution thresholds LGTM 👍

nvtaveras (approved):
Great work 🚀

nvtaveras (changes requested):
One last TODO on my end before merging: I'll check w/Nadiem tomorrow morning if we have values for the XOF smoothing factors and if not, I'll remove the piece of code that sets them 0

script/upgrades/eXOF/eXOF.sol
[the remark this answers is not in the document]
nvtaveras: checked w/ Nadiem, we won't have a smoothing factor for the CELOXOF, as it's similar to other CELOXXX pairs where we also don't have one. I'm removing this tx.

script/upgrades/eXOF/Config.sol
```diff
@@ -0,0 +1,203 @@
...
+  function CELOXOF_RateFeedConfig(
+    Contracts.Cache storage contracts
+  ) internal view returns (Config.RateFeed memory rateFeedConfig) {
+    rateFeedConfig.rateFeedID = contracts.deployed("StableTokenXOFProxy");
+    rateFeedConfig.medianDeltaBreaker0 = Config.MedianDeltaBreaker({
+      enabled: true,
+      threshold: FixidityLib.newFixedFraction(3, 100), // 0.03
+      cooldown: 30 minutes,
+      smoothingFactor: 0
+    });
```
sissnad, line 64:
```suggestion
     });
    rateFeedConfig.dependentRateFeeds = Arrays.addresses(contracts.dependency("EUROXOFRateFeedAddr"));
```

sissnad (changes requested):

script/upgrades/eXOF/Config.sol
nvtaveras: did this change in to also add the deployment checks https://github.com/mento-protocol/mento-deployment/pull/115/commits/43d6531699733a363529496b401a0d2f34d6ade9

nvtaveras: mistakenly used `EUROCEUR` as the dependency while it should be `EURXOF`, fixed here https://github.com/mento-protocol/mento-deployment/pull/115/commits/da68e08b850b90e27ad0303cf29543391605ff2b

nvtaveras (approved):

sissnad (approved):
Looks good!

PR #124, opened 2023-09-04
fix: verify-bytecode script for MU03

nvtaveras:
### Description

From Bogdan on Discord:

> we have these foundry profiles: https://github.com/mento-protocol/mento-deployment/blob/main/foundry.toml#L37-L53 which are meant to reuse the libraries we have already deployed before, and these are included in the metadata.json file as well. But the "deploy" script we used didn't set the correct environment variable, so the metadata had a diff hash

> I fixed the script so this shouldn't happen anymore 
I also just regenerate the cgp json with the new addresses, and because I did this, also included DiWi in the script to be consistent

### Other changes

- Added the changes to Whitelist Diwu's EUROC/EUR oracle + deployment checks

### Tested

- `yarn verify:bytecodes -n celo -u MU03 -c v2.2.0` is green for MU03
- Simulated the MU03 proposal on Mainnet

### Related issues

- Fixes https://github.com/mento-protocol/mento-general/issues/301


bowd (approved):

script/upgrades/MU03/MU03.sol
```diff
@@ -120,7 +120,7 @@ contract MU03 is IMentoUpgrade, GovernanceScript {
 
     vm.startBroadcast(Chain.deployerPrivateKey());
     {
-      createProposal(_transactions, "MU03", governance);
+      createProposal(_transactions, "https://google.com", governance);
```
baroooo, line 123: I assume this will be updated later when we have a descriptionURL?

script/upgrades/MU03/MU03.sol
```diff
@@ -557,6 +558,18 @@ contract MU03 is IMentoUpgrade, GovernanceScript {
     );
   }
 
+  function proposal_whitelistMissingOracleProvider(MU03Config.MU03 memory config) public {
+    // Add DiWu as oracle provider for the EUROC/EUR rate feed
+    address diwuOracleAddress = 0xBD136a625299A0ac5Ca7Ce9220aCA6e08a624e37;
```
baroooo, line 563: where can I verify this ?

nvtaveras: This is the message where he shared his address https://discord.com/channels/966739027782955068/1044219395553120276/1142131663397208094

script/upgrades/MU03/MU03.sol
ninocomputer: yes @nvtaveras

baroooo (approved):

PR #126, opened 2023-09-07
EUROCXOF Rate Feed

sissnad:
### Description

This PR adds EUROCXOF ratefeed config. 

### Other changes

_Describe any minor or "drive-by" changes here._

### Tested

_An explanation of how the changes were tested or an explanation as to why they don't need to be._

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

_Brief explanation of why these changes are/are not backwards compatible._

### Documentation

_The set of community facing docs that have been added/modified because of this change_


sissnad:
We could think about adding a value delta breaker without auto recovery similar to EURXOF. But it feels over-protective as EUROCEUR and EURXOF breakers are rate dependecies

PR #127, opened 2023-09-08
Update eXOF proposal with EUROCXOF rateFeed 

bowd:
### Description

We decided to differentiate between EURXOF (pure exchange rate) and EUROCXOF (EUROC->EUR->XOF) rate which takes into account the EUROCEUR rate as well, and use the latter as the referace rate for the `eXOF/axlEUROC` pool.
This allows us to configure everything using a single ValueDeltaBreaker, so we can remove the nonrecoverable one, as the current one on mainnet has defaultCooldown = 0 as well.

⚠️ Can't fully test this until we have oracles reporting for EUROCXOF ⚠️ 

### Other changes

- Added an `eXOFRevert.sol` which was used to revert the previously passed proposal. This will be updated to reflect the current proposal after it's passed on baklava.
- Changed the `yarn cgp` script to support `-r` for revert.

### Tested

Not fully tested as there are no oracle reports on baklava for EUROCEUR.

### Related issues

Fixes https://github.com/mento-protocol/mento-general/issues/313

### Backwards compatibility

N/A

### Documentation

N/A


script/upgrades/eXOF/Config.sol
```diff
@@ -60,32 +62,37 @@ library eXOFConfig {
...
    */
-  function EURXOF_RateFeedConfig(
+  function EUROCXOF_RateFeedConfig(
     Contracts.Cache storage contracts
   ) internal returns (Config.RateFeed memory rateFeedConfig) {
-    rateFeedConfig.rateFeedID = contracts.dependency("EURXOFRateFeedAddr");
+    rateFeedConfig.rateFeedID = Config.rateFeedID("EUROCXOF");
     rateFeedConfig.valueDeltaBreaker0 = Config.ValueDeltaBreaker({
       enabled: true,
-      threshold: FixidityLib.newFixedFraction(5, 1000), // 0.005
```
sissnad, line 77: Removing this value breaker on EURXOF has two consequences:
1. Only one of the base pairs that build EUROC/XOF have a tight (50bps) value breaker
2. The CELO/eXOF pool will only inherit the 10% value breaker. In particular, CELO/eXOF will stay active while EUROC/eXOF is suspended until it auto-recovers.

I think we could live with 1, by maybe lowering EUROC/XOF value breaker threshold to 50bps. 
But 2 does not feel great. Allowing minting / redemption through CELO/eXOF while we do not via EUROC

script/upgrades/eXOF/Config.sol
```diff
@@ -60,32 +62,37 @@ library eXOFConfig {
...
+      threshold: FixidityLib.newFixedFraction(10, 1000), // 0.01
```
nvtaveras, line 79: Why did we changed this threshold from `0.005` to `0.01`?

sissnad: I would also vote for `0.005`

script/upgrades/eXOF/Config.sol
```diff
@@ -54,38 +56,39 @@ library eXOFConfig {
...
     rateFeedConfig.rateFeedID = contracts.deployed("StableTokenXOFProxy");
     rateFeedConfig.medianDeltaBreaker0 = Config.MedianDeltaBreaker({
       enabled: true,
       threshold: FixidityLib.newFixedFraction(3, 100), // 0.03
       cooldown: 30 minutes,
-      smoothingFactor: 0
+      smoothingFactor: 1e24 // 1
     });
-    rateFeedConfig.dependentRateFeeds = Arrays.addresses(contracts.dependency("EURXOFRateFeedAddr"));
+    rateFeedConfig.dependentRateFeeds = Arrays.addresses(Config.rateFeedID("EURXOF"));
```
sissnad, line 67:
```suggestion
    rateFeedConfig.dependentRateFeeds = Arrays.addresses(Config.rateFeedID("EUROCXOF"));
```

sissnad: With the `EUROCXOF` dependency we should see the followoing behaviour:

1. `Scenario: |EUROC/EUR -1| > 0.005 & EUR/XOF  = 655.597`  => `CELO/eXOF` pool is suspended
2. `Scenario: |EUR/XOF -655.597| > 0.005 & EUROC/EUR =1`  => `CELO/eXOF` pool is suspended
3. `Scenario: EUROC/EUR * EUR/XOF = 655.597 & ( |EUROC/EUR - 1| > 0.005 & |EUR/XOF - 655.597| > 0.005 )` => `CELO/eXOF` is not suspended

sissnad: Just adding this comment to make sure we are on the same page with regard to setting & effects.
1. Is an unnecessary suspension but not a problem
2. Is as planned
3. Is the a weakness of not breaking `EUR/XOF`. But I think it is acceptable to start with

script/upgrades/eXOF/Config.sol
nvtaveras: updated

script/upgrades/eXOF/Config.sol
nvtaveras: As discussed offline, updated to have both `EUROC/XOF` and `EUR/XOF` as dependencies

sissnad (approved):
Good to go 🚀

nvtaveras (approved):
🚢 ⛰️

PR #139, opened 2023-10-31
feat: governance script + deployment checks

### LINKED ISSUE #114: Create MU04 governance script
**Description**
This ticket is about writing the MU04 governance script. The Mento upgrade 4 consist of two main workflows:
1. StableToken Contract upgrade [notion page](https://www.notion.so/Upgraded-StableToken-bc5c334ab7f34b0b923af96cca7e2dcf)
2. deprecation of Mento V1 [notion page](https://www.notion.so/Transition-fully-to-Mento-v2-dc804c0742af422cb02b95f3ad2c20e7)
 
**1) In order to upgrade existing stables to StableTokenV2 we need to execute a few transactions in our gov script:**

- [x] Call `initializeV2()` on the upgraded contracts
- [x] Update StableToken proxies with the new implementations

**2) For the deprecation of Mento V1 we need to execute the following transactions:**

Exchanges:
- [x]  freeze all three Exchanges
- [x]  point exchange identifiers in the Registry to the zero address for all three exchanges:

Reserve:
- [x]  Link reserve Proxy to updated implementation also used for the PartialReserveProxy
- [x]  remove ExchangeEURProxy & ExchangeBRLProxy from exchangeSpender list
- [x]  add broker to exchangeSpender list
- [x]  configure collateral assets:
    - [x]  add axlEUROC as a collateral asset in the main reserve `addCollateralAsset`
    - [x]  set axlEUROC spending ratio in the main reserve to ??
    - [x]  add axlUSDC as a collateral asset in the main reserve `addCollateralAsset`
    - [x]  set axlUSDC spending ratio in the main reserve to ??
    - [x]  add Celo as a collateral asset in the main reserve `addCollateralAsset`
    - [x]  set Celo spending ratio in the main reserve to ??
- [x]  add new stable assets:
    - [x]  add eXOF as a token `addToken`
- [x]  add partial reserve multisig as a spender `addSpender`
- [x]  Remove unnecessary spenders from the main reserve

Broker:
- [x]  call `setReserve` on Broker with address of main reserve as param
- [x]  set TradingLimits for Pools if changed

BiPoolManager:
- [x]  call `setReserve` on BiPoolManager with the address of the main reserve as the param
- [ ]  redeploy pools if any parameters have changed e.g Bucket Size

**Acceptance Criteria:**
- [x] Governance Script is written
- [x] Checks are written 
- [x] Add a readme about the upgrade

philbow61:
### Description

This PR adds the governance script for MU04 and the corresponding deployment checks.
as of now the script only works on baklava since the StableTokenV2 has only been deployed to baklava.
There are a couple of things to note:
- the config comes from @sissnad  and includes pools where the config hasn't changed. This is done since the changes of MU04 affect all pools and this way it's easier to verify swaps still work after the upgrade (checkout: MU04Checks.swap.sol). This means the MU04 config represents the config every pool should have after the MU04 upgrade was executed.   
- In MU04 we want to add the PartialReserveMultisig as the only ExchangeSpender of the main Reserve. Therefore we need to remove existing reserve spenders. Since the Reserve doesn't expose a function that lets you retrieve all spenders I used a dune script to query all transactions that added spenders  https://dune.com/queries/3086623. Based on the result I believe there is only one reserve spender currently(0x554Fca0f7c465cd2F8C305a10bF907A2034d2a19). That is why the address is hardcoded like this throughout the repo `oldMainReserveMultisig = 0x554Fca0f7c465cd2F8C305a10bF907A2034d2a19;`. Since this is only true for Mainnet I added a check for this address being a spender before removing it. I think it's fine if we only make this change to the reserveSpender list on Mainnet. (not super exited to figure out how to get current exchangeSpenders on baklava/alfajores 😁  )
- the spendingRatios for reserve collateral assets are currently hardcoded to 50%. I came up with these numbers in order to test the script and the checks once we have decided on the actual values I will update them.
- added the exchange + GrandaMento Addresses to the dependencies.json not sure if that is the way we want to do it but after removing their registry entries we need a way to access them for each chain if we want to verify those changes :/

Here are the MU04 notion pages for reference:
- https://www.notion.so/Upgraded-StableToken-bc5c334ab7f34b0b923af96cca7e2dcf
- https://www.notion.so/Transition-fully-to-Mento-v2-dc804c0742af422cb02b95f3ad2c20e7

### Other changes
added the IFreezer interface in order to freeze the exchange contracts. 

### Tested

- Simulation is green on baklava run `yarn cgp -n baklava -u MU04 -s ` to verify

### Related issues

- Fixes #114


script/upgrades/MU04/MU04.sol
```diff
@@ -0,0 +1,494 @@
...
+   * @dev Loads the deployed contracts from the previous deployment step
+   */
+  function loadDeployedContracts() public {
+    contracts.load("MU01-00-Create-Proxies", "latest");
+    contracts.load("MU01-01-Create-Nonupgradeable-Contracts", "latest");
+    contracts.load("MU03-01-Create-Nonupgradeable-Contracts", "latest");
+    contracts.load("MU03-02-Create-Implementations", "latest");
+    contracts.load("eXOF-00-Create-Proxies", "latest");
+    contracts.load("eXOF-01-Create-Implementations", "latest");
+    contracts.load("MU04-00-Create-Implementations", "latest");
```
bowd, line 83: Do we need all of these? Just by looking at the `contracts.deployed` calls I think we shouldn't.

script/upgrades/MU04/MU04.sol
```diff
@@ -0,0 +1,494 @@
...
+    for (uint i = 0; i < stableTokens.length; i++) {
+      transactions.push(
+        ICeloGovernance.Transaction(
+          0,
+          stableTokens[i],
+          abi.encodeWithSelector(
+            IStableTokenV2(0).initializeV2.selector,
+            brokerProxy,
+            validators,
+            address(0) // exchange contracts are depricated
```
bowd, line 205: deprecate not depricate 🤓

script/upgrades/MU04/MU04.sol
```diff
@@ -0,0 +1,494 @@
...
+  /**
+   * @dev Updates reserve in BiPoolManager to use the main reserve
+   */
+  function proposal_updateReserveInBiPoolManager() private {
+    if (address(BiPoolManager(biPoolManagerProxy).reserve()) == partialReserveProxy) {
+      transactions.push(
+        ICeloGovernance.Transaction(
+          0,
+          biPoolManagerProxy,
+          abi.encodeWithSelector(Broker(0).setReserve.selector, reserveProxy)
```
bowd, line 384: This works because the function has the same selector, but to remove confusion let's use `BiPoolManager(0).setReserve.selector` here

script/upgrades/MU04/MU04Checks.swap.sol
```diff
@@ -0,0 +1,536 @@
...
+      tokenOut,
+      amountIn,
+      config.cBRLCelo.referenceRateFeedID,
+      FixidityLib.wrap(config.cBRLCelo.spread.unwrap())
+    );
+
+    console.log("\t🟢 CELO -> cBRL swap successful 🚀");
+  }
+
+  function swapcBrlToCELO(MU04Config.MU04 memory config) internal {
```
bowd, line 186: Nit: cBrl -> cBRL

script/upgrades/dependencies.json
```diff
@@ -27,6 +35,10 @@
     "USDCBRLRateFeedAddr": "0x25F21A1f97607Edf6852339fad709728cffb9a9d",
     "EURXOFRateFeedAddr": "0x40dc8528167557353fdcd98548ab2139a670dd0b",
     "EUROCEURRateFeedAddr": "0x26076B9702885d475ac8c3dB3Bd9F250Dc5A318B",
-    "PartialReserveMultisig": "0x56fD3F2bEE130e9867942D0F463a16fBE49B8d81"
+    "PartialReserveMultisig": "0x56fD3F2bEE130e9867942D0F463a16fBE49B8d81",
+    "Exchange": "0x17bc3304f94c85618c46d0888aa937148007bd3c",
+    "ExchangeEUR": "0x997b494f17d3c49e66fafb50f37a972d8db9325b",
+    "ExchangeBRL": "0xf391dcaf77360d39e566b93c8c0ceb7128fa1a08",
+    "GrandaMento": "0xecf09fcd57b0c8b1fd3de92d59e234b88938485b"
```
bowd, line 42: While not pretty I think this is fine.

bowd (changes requested):
Great job on this, left minor comments :shipit:

script/upgrades/MU04/MU04.sol
```diff
@@ -0,0 +1,496 @@
...
+            validators,
+            address(0) // exchange contracts are depricated
+          )
+        )
+      );
+    }
+  }
+
+  /**
+   * @dev Freezes all exchanges in order to depricate Mento V1
```
nvtaveras, line 213:
```suggestion
   * @dev Freezes all exchanges in order to deprecate Mento V1
```

script/upgrades/MU04/MU04.sol
```diff
@@ -0,0 +1,496 @@
...
+          spendingRatios
+        )
+      )
+    );
+  }
+
+  /**
+   * @dev Updates reserve tokens to include eXOF
+   */
+  function proposal_updateReserveTokens() private {
```
nvtaveras, line 329: how about renaming to something like`proposal_addEXOFToMainReserve` ?

script/upgrades/MU04/MU04Checks.verify.sol
```diff
@@ -0,0 +1,323 @@
...
+    return string(bytesArray);
+  }
+
+  function verifyCollateralSpendingRatio(address collateralAsset, uint256 expectedRatio) internal {
+    console.log("\n== Verifying Collateral Spending Ratio for %s ==", collateralAsset);
+    // @notice verifiying spending ratios by trying moving the allowed and more than the allowed amount
+    // the variable holding the ratios is private
+
+    address[] memory otherReserveAddresses = Reserve(reserveProxy).getOtherReserveAddresses();
+    if (otherReserveAddresses.length > 0) {
```
nvtaveras, line 299: is there a network for which this is not true? If not, let's change this to a require instead?

nvtaveras (commented):
Great work, just some minor comments

script/upgrades/MU04/MU04Checks.verify.sol
philbow61: I haven't checked but let me find out and if all of them have another reserve I will remove it 👍

nvtaveras (approved):
🥇

bowd (approved):
📦 🚀

PR #142, opened 2023-12-06
fix: remove isCollateralAsset check in MU04

nvtaveras:
### Description

While simulating MU04 on Alfajores we ran into a revert which was caused due to calling `isCollateralAsset` on `ReserveProxy` that is still pointing to the legacy implementation. Our speculation is that the reason this worked on Baklava is that at some point we upgraded the implementation of the legacy reserve contract to the new one outside of MU04.

### Tested

Simulation now works on Alfajores ✅


philbow61 (approved):
👍

baroooo (approved):

PR #161, opened 2024-02-08
Fix for get-secrets.sh: added macOS-compatible path for `sed` util usage

chapati23:
`yarn secrets:get` errored on my machine with a shell syntax error on the `sed` usage which slightly differs between linux and macOS environments

This PR adds a macOS-specific version of this command (if runtime env is macOS, else nothing changes)

On the `main` branch, running `yarn secrets:get` threw an error:

```
$ ./bin/get-secrets.sh ./bin/../.env sed: 1: "./bin/../.env": invalid command code . sed: 1: "./bin/../.env": invalid command code . sed: 1: "./bin/../.env": invalid command code . error Command failed with exit code 1.
```

Copilot had this to say on the matter which made sense and the syntax change works for me locally:

image

### DIFF
--- a/bin/get-secrets.sh
+++ b/bin/get-secrets.sh
@@ -2,7 +2,7 @@
 
 ##############################################################################
 # Script for pulling secret keys from GCP secret manager.
-# Usage: ./bin/get-secrets.sh 
+# Usage: ./bin/get-secrets.sh
 ##############################################################################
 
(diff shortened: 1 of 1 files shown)


bayological (approved):
Tested this on both MacOS and Ubuntu, both work fine.

baroooo (approved):

PR #162, opened 2024-02-11
feat: transition governance deployment to Forge

bowd:
### Description

When implementing the governance deployment, we hit a few roadblocks, and introduced `hardhat` for this step. While it good thing that it kept us moving forward, it caused this repo to get way more complicated and introduced additional cognitive load to the process. This PR removes that and reimplements the deployment in the existing paradigm by doing the following:

1. Deploy the GovernanceFactory in a `solc-0.8` script.
2. Create a proposal and checks following the existing pattern.
3. Create a new `cgp:execute` command, which uses a foundry script to trigger execution. This has the added benefit of automatically verifying any contracts created as part of that transaction (i.e., all contracts that the Factory creates).
4. Replace the `celocli` command in the `cgp:pass` script with the `cgp:execute` command.
5. Move the merkle tree related scripts into a `bin/` command to keep the repo structure.

This PR adds a `mento-core-2.3.0` submodule, which currently points to a branch containing all the latest changes to the core contracts, but in the end, it will be pinned to the `v2.3.0` tag. That version is used for the Governance MU.

Additionally, I have prepared #163, which makes it easy to verify contracts on `celoscan` as long as they were verified on sourcify. More details in that PR.

### Other changes

N/A

### Tested

Deployments ran.

### Related issues

N/A

### Backwards compatibility

N/A

### Documentation

N/A


bin/verify-celoscan.ts
```diff
@@ -0,0 +1,270 @@
+/*******************************************************************
```
bowd, line 1: This code is adapted from https://github.com/k1rill-fedoseev/sourcify-to-etherscan
And fixes some bugs there

script/upgrades/dependencies.json
```diff
@@ -25,7 +28,10 @@
     "Exchange": "0x190480908c11Efca37EDEA4405f4cE1703b68b23",
     "ExchangeEUR": "0xC200CD8ac71A63e38646C34b51ee3cBA159dB544",
     "ExchangeBRL": "0x28e257d1E73018A116A7C68E9d07eba736D9Ec05",
-    "GrandaMento": "0xdfd641aB188Add84B317fB0b241F6b879E5EF906"
+    "GrandaMento": "0xdfd641aB188Add84B317fB0b241F6b879E5EF906",
+    "MentoLabsMultisig": "0x",
```
baroooo, line 32: we are missing some addresses for baklava

baroooo (approved):
Looks great 🎩 🎩
Ran into some issues with verification but I guess they will be fixed with the next PR

script/upgrades/MUGOV/deploy/MUGOV-04-Create-Factory.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
baroooo, line 1: we started from 04 :)

baroooo: I mean the NR of the script is 04

script/upgrades/MUGOV/MUGOVChecks.sol
```diff
@@ -0,0 +1,144 @@
...
+  function run() public {
+    // Balances:
+    assertEq(mentoToken.balanceOf(mentoLabsMultisig), 100000000 * 1e18, "MentoLabs multisig balance is incorrect");
+    assertEq(mentoToken.balanceOf(address(airgrab)), 100000000 * 1e18, "Airgrab balance is incorrect");
+    assertEq(
+      mentoToken.balanceOf(address(governanceTimelock)),
+      100000000 * 1e18,
+      "Governance timelock balance is incorrect"
+    );
+    assertEq(mentoToken.emissionSupply(), 693000000 * 1e18, "Emission supply is incorrect");
```
baroooo, line 60: some of these checks won't pass on mainnet.
we can come back to this laster if the allocations are not final

philbow61 (approved):
💯 
I Went through the gov process on Alfajores and everything worked smoothly 👍

PR #163, opened 2024-02-13
feat: add command to verify contracts on celoscan

bowd:
### Description

We need our contracts to be verified both on Sourcify (for Celo's Blockscout) and Celoscan. 
Sourcify works out of the box with `foundry` but there were problems getting `celoscan` to work well in the past.
This script strips out parts of the [sourcify-to-etherscan](https://github.com/k1rill-fedoseev/sourcify-to-etherscan) repo, fixes some issues and adds some extra logic to compute `constructor-args` on the fly when it can.

The command works by running:
```
$ yarn verify:celoscan ./broadcast/MUGOV-04-Create-Factory.sol/44787/run-latest.json
```

It will go through the broadcast file and extract any contract creations. These can be either direct from the script or as a result of calling a factory contract (like we're doing for Governance).

It will extract the init code and compare it with the compiled bytecode and determine constructor args, build the library list correctly and push that to celoscan.

### Other changes

N/A

### Tested

Yes.

### Backwards compatibility

N/A

### Documentation

N/A


bin/verify-celoscan.ts
```diff
@@ -0,0 +1,267 @@
...
+    if (initCode.length < bytecode.length) {
+      console.error("🚨 contract creation data is shorter than bytecode, something is off");
+      return "";
+    }
+    if (initCode.length == bytecode.length) {
+      console.error(" -- contract creation data is equal to bytecode, assuming no constructor args");
+      return "";
+    } else {
+      console.log(" -- contract creation data is longer than bytecode, assuming constructor args");
+      const diff = initCode.slice(bytecode.length);
```
baroooo, line 212: 🧠

bin/verify-celoscan.ts
```diff
@@ -0,0 +1,267 @@
...
+const fileContent = fs.readFileSync(broadcastFile, "utf8");
+const broadcast = JSON.parse(fileContent) as BroadcastFile;
+
+let chain: "alfajores" | "celo" | null = null;
+let celoscanApiUrl: string;
+let celoscanApiKey: string;
+if (broadcast.chain == 44787) {
+  chain = "alfajores";
+  celoscanApiUrl = "https://api-alfajores.celoscan.io/api";
+  if (!process.env.CELOSCAN_ALFAJORES_API_KEY) {
```
baroooo, line 61: same key should work for all testnets

baroooo (approved):

bayological (approved):

PR #164, opened 2024-02-28
feat: MU05 script + deployment checks

### LINKED ISSUE #155: Create MU06 proposal and checks
**Description** 

> Check Epic for context

This task covers the creation of the MU06 governance proposal script which would replicate all axlUSDC pools but with the native USDC on Celo.

**Acceptance criteria** 

- [ ] Proposal is drafted and tested via simulation

philbow61:
### Description

This PR adds the deployment script + deployment checks for native USDC pools(MU05)
In the ticket, it's called MU06. I am not sure if I am missing something but since there is no MU05 so far I called it MU05.

The new pools use the same configs as the existing bridged USDC pools. But as Nadiem pointed out we use half the global limit for the non USD to USD pairs to keep our FX rate risk the same 

### Other changes
- deployed a mock ERC20 native USDC to testnets 

### Tested

- simulation passes on all networks 

### Related issues

- Fixes #155


script/upgrades/MU05/Config.sol
```diff
@@ -0,0 +1,112 @@
...
+      referenceRateFeedID: contracts.dependency("USDCEURRateFeedAddr"),
+      asset0limits: Config.TradingLimit({
+        enabled0: true,
+        timeStep0: 5 minutes,
+        limit0: 500_000,
+        enabled1: true,
+        timeStep1: 1 days,
+        limit1: 1_000_000,
+        enabledGlobal: true,
+        limitGlobal: 14_000_000
```
sissnad, line 75:
```suggestion
        limitGlobal: 7_000_000
```

script/upgrades/MU05/Config.sol
```diff
@@ -0,0 +1,112 @@
...
+      referenceRateFeedID: contracts.dependency("USDCBRLRateFeedAddr"),
+      asset0limits: Config.TradingLimit({
+        enabled0: true,
+        timeStep0: 5 minutes,
+        limit0: 500_000,
+        enabled1: true,
+        timeStep1: 1 days,
+        limit1: 1_000_000,
+        enabledGlobal: true,
+        limitGlobal: 5_000_000
```
sissnad, line 103:
```suggestion
        limitGlobal: 2_500_000
```

sissnad (changes requested):
I would suggest to set the global limits to half of the current global limit. The idea is to keep the total limit that we agreed before on how much FX risk can be created. 
The global limits for the corresponding axlUSD pools should also be reduced by a factor 0.5.

script/upgrades/MU05/MU05Checks.verify.sol
```diff
@@ -0,0 +1,127 @@
...
+      checkTradingLimt(poolConfig.asset0limits, asset0ActualLimit);
+      checkTradingLimt(poolConfig.asset1limits, asset1ActualLimit);
+    }
+
+    console.log("🟢 Trading limits correctly updated for all exchanges 🔒");
+  }
+
+  function verifyCollateralSpendingRatio(address collateralAsset, uint256 expectedRatio) internal {
+    console.log("\n== Verifying Collateral Spending Ratio for %s ==", collateralAsset);
+    // @notice verifiying spending ratios by trying moving the allowed and more than the allowed amount
```
chapati23, line 74:
```suggestion
    // @notice verifying spending ratios by trying to move the allowed, and more than the allowed amount
```

chapati23 (approved):
lgtm! found just one typo but no blockers

double-checked correct USDC address on celo against this https://www.circle.com/blog/usdc-now-available-on-celo

script/upgrades/MU05/MU05.sol
```diff
@@ -0,0 +1,215 @@
...
+      transactions.push(
+        ICeloGovernance.Transaction(
+          0,
+          reserveProxy,
+          abi.encodeWithSelector(Reserve(0).addCollateralAsset.selector, nativeUSDC)
+        )
+      );
+    }
+
+    // Set native USDC daily spending ratio to 100% sames as axlUSDC
```
nvtaveras, line 129:
```suggestion
    // Set native USDC daily spending ratio to 100%, same as axlUSDC
```

script/upgrades/MU05/MU05.sol
```diff
@@ -0,0 +1,215 @@
...
+        )
+      );
+    }
+  }
+
+  /**
+   * @notice This function creates the transactions to create the new pairs with native USDC.
+   */
+  function proposal_createExchanges(MU05Config.MU05 memory config) private {
+    // Get the address of the pricing modules
```
nvtaveras, line 149: redundant?

script/upgrades/MU05/MU05.sol
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  /**
+   * @notice This function creates the transactions to configure the trading limits.
+   */
+  function proposal_configureTradingLimits(MU05Config.MU05 memory config) public {
+    for (uint256 i = 0; i < config.pools.length; i++) {
+      Config.Pool memory poolConfig = config.pools[i];
+
+      // Set the trading limits for the pool
```
nvtaveras, line 193: also redundant?

script/upgrades/dependencies.json
```diff
@@ -35,6 +37,7 @@
   },
   "44787": {
     "BridgedUSDC": "0x87D61dA3d668797786D73BC674F053f87111570d",
+    "NativeUSDC": "0x17dAe16f0EcB55FCd6887b01C59aD1F5e592d490",
```
nvtaveras, line 40: Pending to see if we can replace this by the actual USDC deployed on Alfajores https://developers.circle.com/stablecoins/docs/usdc-on-test-networks

script/upgrades/MU05/MU05Checks.verify.sol
```diff
@@ -0,0 +1,127 @@
...
+
+  function verifyReserveCollateralAssets() internal {
+    console.log("\n== Verifying Reserve Collateral Assets ==");
+
+    require(Reserve(reserveProxy).isCollateralAsset(nativeUSDC), "❗️❌ Reserve collateral asset not set correctly");
+
+    // mint some native USDC to the main reserve in order to verify spending ratios
+    deal(nativeUSDC, reserveProxy, 100e6);
+    verifyCollateralSpendingRatio(nativeUSDC, 1e24);
+    console.log("🟢 Asset: %s successfully added to collateral asset list", nativeUSDC);
```
nvtaveras, line 48: move this below the require statement in L43?

script/upgrades/MU05/MU05Checks.verify.sol
```diff
@@ -0,0 +1,127 @@
...
+    // the variable holding the ratios is private
+
+    address[] memory otherReserveAddresses = Reserve(reserveProxy).getOtherReserveAddresses();
+    address payable otherReserve = address(uint160(otherReserveAddresses[0]));
+    uint256 reserveBalance = Reserve(reserveProxy).getReserveAddressesCollateralAssetBalance(collateralAsset);
+
+    uint256 spendingLimit = FixidityLib.wrap(expectedRatio).multiply(FixidityLib.newFixed(reserveBalance)).fromFixed();
+    uint256 exceedingAmount = spendingLimit + 1;
+
+    vm.prank(partialReserveMultisig);
```
nvtaveras, line 84: was initially confused on why we needed this after we moved away from the partial reserve. What do you think of renaming to `reserveSpender`?

nvtaveras (approved):
LGTM, just some minor comments

script/upgrades/MU05/MU05Checks.verify.sol
philbow61: the idea was that it is successfully added if it is both a collateral asset and the spending ratio is set correctly.

script/upgrades/MU05/MU05Checks.verify.sol
philbow61: i agree sounds better 👍

nvtaveras (approved):

sissnad (approved):

PR #167, opened 2024-04-08
Feat/new root and fractal signer

baroooo:
### Description

In order to make airgrab integration tests work, we need to update the merkle tree to include one of the whitelisted accounts and use the actual fractal signer to verify signatures.
This pr makes necessary changes and deploys the testnet contracts

### Other changes

- Removed hardhat config as it is not used anymore
- Updated governance factory interface to match the last implementation
- Updated mappings
- Fixed a typo on deployment checks

### Tested

Deployment was successful
Post deployment checks are green
Ran a simulated claim using fractal proof and signer

### Related issues

Required for Add integration tests for Airgrab contract#3


philbow61 (approved):
LGTM but we always just pushed the run-lattest.json not all broadcast files

nvtaveras (approved):

baroooo:
> LGTM but we always just pushed the run-lattest.json not all broadcast files

alright, removed older broadcasts

PR #168, opened 2024-04-17
feat: final gov deployment allocations and merkle tree

nvtaveras:
### Description

This PR partially completes https://github.com/mento-protocol/mento-core/issues/405 as we currently don't have all the final addresses for each allocation, so not all the addresses are final (the ones that need to be updated have been marked with a TODO).

The following however has been updated:

1. Final token allocation according to the [forum post](https://forum.celo.org/t/mento-spin-off-and-launch-of-the-mento-token/7747)

2. ~Merkle tree generated from the final snapshot (`airdrop-amounts-per-address.csv` in https://github.com/mento-protocol/airdrop-snapshot/tree/main/final-snapshots)~
**update: this is no longer the final snapshot as there are still some modifications being made. I have also replaced one of the recipients in the original CSV with one of our test accounts which we'll use in the integration tests**

4. ~Fractal signer address on Alfajores to now use the mainnet signer~ 
**reverted: we need the staging signer for the integration tests CI**

Other changes left to do:
- [x] Update airgrab duration (waiting for confirmation from @olenovyk)

### Other changes

I made each allocation address unique (even if temporary) so that we can verify that there's no address mix-up. Reach out if you need the private key for any of them for testing purposes.

### Tested

Ran a simulation of the deployment on Alfajores


script/upgrades/MUGOV/MUGOVChecks.sol
```diff
@@ -86,9 +98,12 @@ contract MUGOVChecks is GovernanceScript, Test {
     assertEq(airgrab.token(), address(mentoToken), "Airgrab token is incorrect");
     assertEq(airgrab.locking(), address(locking), "Airgrab locking is incorrect");
     assertEq(airgrab.mentoTreasury(), address(governanceTimelock), "Airgrab Mento Treasury is incorrect");
+
+    // @TODO: Replace below check with 2 weeks after updating mento-core
+    assertEq(airgrab.endTimestamp() - block.timestamp, 365 days, "Airgrab end timestamp is incorrect");
```
baroooo, line 103: I am not sure if the strict equality will hold here, assuming the checks will run one block after the deployment

nvtaveras: both the deployment + checks all happen on the same block

baroooo (approved):

philbow61 (approved):
💯

PR #195, opened 2024-05-23
Feat/fix mento core versioning

bowd:
### Description

We had a bit of a versioning weirdness in the interim between cutting new versions. We now have cut these two new versions on mento-core:

v2.3.0 - the final audit commit
v2.3.1 - adds the new stable token proxies + changes the threshold in the GovernanceFactory

I've updated mento deployment to remove the linked versions v2.3.0 and v2.4.0 and replace them with v2.3.1 and point all references to that.

### Other changes

N/A

### Tested

Ran locally

### Related issues

N/A

### Backwards compatibility

N/A

### Documentation

N/A

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "lib/mento-core-2.4.0"]
 	path = lib/mento-core-2.4.0
 	url = https://github.com/mento-protocol/mento-core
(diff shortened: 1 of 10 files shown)


philbow61 (approved):
LGTM

PR #197, opened 2024-07-15
feat: update bash scripts to work for both governances

baroooo:
### Description

Updates bash scripts to work with both Celo and Mento governances at the same time by introducing a new flag "-g".
If the governance is "celo", everything should work as before
If the governance is "mento", util scripts under "/mento" subfolder should be used to perform operations on the Mento Governance

### Other changes

Removed some unused flags and updated some related comments in bash scripts

### Tested

Tested by running MUGOV on Celo gov
Mento gov side can not be tested until the rest of the scripts are created

### Related issues

- Fixes [#476](https://github.com/mento-protocol/mento-general/issues/476)

### Backwards compatibility

Compatible

### DIFF
--- a/bin/cgp-check.sh
+++ b/bin/cgp-check.sh
@@ -5,23 +5,27 @@
 # Usage: yarn cgp:check
 #               -n <baklava|alfajores|celo>  -- network to submit the proposal to
 #               -u <upgrade_name>            -- name of the upgrade (MU01)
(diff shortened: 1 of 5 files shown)


nvtaveras (approved):
LGTM

bayological (approved):
LGTM

chapati23 (approved):
lgtm 👍

PR #198, opened 2024-07-19
feat: mento gov utils scripts

baroooo:
### Description

Implements the scripts to create and execute proposals on Mento Governance

### Other changes

- updates pragma of files used by two governances to a accept both solc versions
- adds bravo compatible interface to interact with the mento governance

### How to review

Get a diff between scripts under `script/utils/mento` and `script/utils` and confirm if the changes correctly reflects the differences between celo and mento governances

### Tested

By running MUGOV on Mento Governance. 
https://alfajores.celoscan.io/tx/0x5ca16d5ef582a6ce2133735db348372b1d796e99bef647818a499d294d53203a
GovernanceFactory: 0x9491cF9e591BD3dce3936431d8563529B5024faD

Note: Blockscout is not indexing new contracts at this moment. For instance the newly deployed airdrop contract is not indexed on blockscout but celoscan was able to pick it up:
https://explorer.celo.org/alfajores/address/0x8958403666CfeE018C20ecE0eBE9B152260cCbfc
https://alfajores.celoscan.io/address/0x8958403666CfeE018C20ecE0eBE9B152260cCbfc#code
Also not all blocks are getting indexed for a while: https://explorer.celo.org/alfajores/blocks
I believe contract verification fails because of this reason.

You can perform a similar test with a couple adjustments. How I was able to test:

- Updated `MUGOV-00-Create-Factory.sol` to use the TimelockController as the owner
- Updated `MUGOV.sol` to use util scripts under `script/utils/mento`, updated solc version used, re applied `using` statement Contracts.Cache and updated governance address as MentoGovernor
- Updated `MUGOVChecks.sol` to use util scripts under `script/utils/mento`, updated solc version used, re applied `using` statement Contracts.Cache, removed `PrecompileHandler` and used `unicode` prefix for string with unicode
- `forge clean`
- `forge build`
- `yarn deploy ...`
- `yarn cgp ...`
- `castVote` and `queue` manually using the deployer account on celoscan
- `yarn cgp:execute ...`

### Related issues

- Fixes [#474](https://github.com/mento-protocol/mento-general/issues/474)

### Backwards compatibility

After all changes, I re-ran the MUGOV using the Celo governance and it worked as before
https://alfajores.celoscan.io/tx/0x2d1201b900e3e45f1eb79d58a6ac4b766ab0b14ad41c459da6364ac3ab2bfef9

### DIFF
--- a/script/utils/mento/Contracts.sol
+++ b/script/utils/mento/Contracts.sol
@@ -1,0 +1,151 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.18;
+
+import { console } from "forge-std/console.sol";
+
+import { Vm } from "forge-std/Vm.sol";
+import { Chain } from "./Chain.sol";
+import { stdJson } from "forge-std/StdJson.sol";
+import { IRegistry } from "script/interfaces/IRegistry.sol";
+
+library Contracts {
+  using stdJson for string;
+
+  address private constant REGISTRY_ADDRESS = 0x000000000000000000000000000000000000ce10;
+  // solhint-disable-next-line const-name-snakecase
+  IRegistry private constant registry = IRegistry(REGISTRY_ADDRESS);
+
+  address private constant VM_ADDRESS = address(bytes20(uint160(uint256(keccak256("hevm cheat code")))));
+  // solhint-disable-next-line const-name-snakecase
+  Vm private constant vm = Vm(VM_ADDRESS);
+
+  bytes32 private constant CREATE_HASH = 0x14feaf0665b390ef0561125746780cd06c1876ebed7339648fad78cd5fb754ed;
+
+  struct Cache {
+    mapping(bytes32 => address) contractAddress;
+    bool _dependenciesLoaded;
+    string _dependencies;
+  }
+
+  function loadUpgrade(Cache storage self, string memory upgrade) internal {
+    load(self, string(abi.encodePacked(upgrade, "-00-Create-Proxies")), "latest");
+    load(self, string(abi.encodePacked(upgrade, "-01-Create-Nonupgradeable-Contracts")), "latest");
+    load(self, string(abi.encodePacked(upgrade, "-02-Create-Implementations")), "latest");
+  }
+
+  function load(Cache storage self, string memory script, string memory timestamp) internal {
+    _load(self, script, timestamp, false);
+  }
+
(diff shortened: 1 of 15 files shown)


script/utils/mento/Contracts.sol
```diff
@@ -0,0 +1,151 @@
...
+
+  function _load(Cache storage self, string memory script, string memory timestamp, bool silent) internal {
+    string memory chainId = Chain.idString();
+    string memory root = vm.projectRoot();
+    string memory path = string(
+      abi.encodePacked(root, "/broadcast/", script, ".sol/", chainId, "/", "run-", timestamp, ".json")
+    );
+    string memory json = vm.readFile(path);
+    /**
+     * note(bogdan): Decoding this isn't straightforwad because we're using
```
philbow61, line 52: Is this comment still necessary for the newer solidity version?

script/utils/mento/GovernanceHelper.sol
```diff
@@ -0,0 +1,76 @@
...
+      govTx.values[i] = transactions[i].value;
+      govTx.calldatas[i] = transactions[i].data;
+    }
+
+    uint256 proposalId = IGovernor(governance).propose(govTx.targets, govTx.values, govTx.calldatas, govTx.description);
+
+    console2.log("Proposal was successfully created. ID: ", proposalId);
+  }
+
+  function simulateProposal(ICeloGovernance.Transaction[] memory transactions, address governance) internal {
```
philbow61, line 47: why are the transactions here of type ICeloGovernance.Transaction?

philbow61 (approved):
LGTM. 
Was able to simulate and propose a proposal for Mento Governance on the new solidity version 👍

script/utils/mento/Contracts.sol
baroooo: I did not touch the way how we load the contracts so we need to update this function and start using mentioned helper scripts instead to remove the comment

script/utils/mento/GovernanceHelper.sol
baroooo: the same tx structure is kept for the new governance so that we can keep writing the MUxx scripts the same way we used to. the main idea was to be able to use the repo for both governances with minimal changes and this structure allows us to use MUs and helper functionalities around that with both governances.

nvtaveras (approved):
🚢

PR #204, opened 2024-08-08
chore: reproduce deployment broadcast files from on-chain data

bowd:
### Description

When I deployed and executed the governance proposal for Mento Governance, I made the immense mistake of not committing the broadcast files to Github. A few days later my macbook went belly up and I lost all the data. Sigh. 

This PR uses an archive node (quicknode) and some manual data to reconstruct the Foundry broadcast files. This is useful to have in conjunction with the rest of our tooling which depends on broadcast files. It would have been very useful a few months ago because it makes it really easy to verify contracts on `celoscan`, I used it to verify a few remaining proxies. 

The script to generate this is a one-off script (famous last words), but I thought it made sense to put it in there in case we ever need something similar, but it shouldn't really be reviewed thoroughly.

### Other changes

N/A

### Tested

Verified broadcast script manually and tried it out with the cleoscan verification script.

### Related issues

N/A

### Backwards compatibility

Precisely 

### Documentation

N/A


bin/reproduce-mainnet-gov.ts
```diff
@@ -0,0 +1,292 @@
+/*******************************************************************
+ * One-off script used to reproduce the mainned Mento Governance
```
RyRy79261, line 2:
```suggestion
 * One-off script used to reproduce the mainnet Mento Governance
```

RyRy79261 (approved):

bin/reproduce-mainnet-gov.ts
```diff
@@ -0,0 +1,292 @@
...
+    contractAddress: trace["result"]["to"],
+    function: null,
+    arguments: null,
+    transaction: {
+      type: tx["typeHex"],
+      from: tx["from"],
+      gas: `0x${tx["gas"].toString(16)}`,
+      value: `0x${tx["value"].toString(16)}`,
+      data: trace["result"]["input"],
+      nonce: tx["nonce"]
```
baroooo, line 257:
```suggestion
      nonce: `0x${tx["nonce"].toString(16)}`
```

baroooo (approved):
Added slight modifications to reduce the differences between the foundry generated and script generated broadcast files. Not a blocker for this, feel free to re-gen broadcast files if you see a need.

bin/reproduce-mainnet-gov.ts
```diff
@@ -0,0 +1,292 @@
...
+    const entry = {
+      hash: dt.txHash,
+      transactionType: trace["result"]["type"],
+      contractName: dt.contractName,
+      contractAddress: trace["result"]["to"],
+      function: null,
+      arguments: dt.arguments,
+      transaction: {
+        type: tx["typeHex"],
+        from: tx["from"],
```
baroooo, line 163:
```suggestion
        from: tx["from"],
        to: tx["to"],
```

PR #207, opened 2024-08-12
PUSO (PSO) config

rcroessmann:
Context:

Naming is a bit unusual: Stablecoin is supposed to be named “PUSO'' which means “heart” in Filipino. Ticker is PSO and reference rate is the Philippine Peso (PHP). 

PHP/USD spread in fx markets are quite tight and chainlink deviation threshold (DT) for the PHP/USD feed is 0.3% - suggested 0.3% spread therefore with the intend to lower it when chainlink lowers the DT.

Trading limits 2x cKES - that means 5m in USD max. supply cap for now.

Same circuit breaker parameters as cKES.

### DIFF
--- a/script/upgrades/PUSO/Config.sol
+++ b/script/upgrades/PUSO/Config.sol
@@ -31,7 +31,7 @@ library PSOConfig {
   function get(Contracts.Cache storage contracts) internal view returns (PSO memory config) {
     config.poolConfig = PSOcUSD_PoolConfig(contracts);
     config.rateFeedConfig = PHPUSD_RateFeedConfig();
-    config.stableTokenConfig = stableTokenKESConfig();
+    config.stableTokenConfig = stableTokenPSOConfig();
   }
 
   /* ==================== Rate Feed Configuration ==================== */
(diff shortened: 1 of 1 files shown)


script/upgrades/PUSO/Config.sol
```diff
@@ -0,0 +1,104 @@
...
+      poolConfig.minimumReports = 1;
+    }
+  }
+
+  /* ==================== Stable Token Configuration ==================== */
+
+  /**
+   * @dev Returns the configuration for the PUSO stable token.
+   */
+  function stableTokenKESConfig() internal pure returns (Config.StableTokenV2 memory config) {
```
bayological, line 101:
```suggestion
  function stableTokenPSOConfig() internal pure returns (Config.StableTokenV2 memory config) {
```

script/upgrades/PUSO/Config.sol
```diff
@@ -0,0 +1,104 @@
...
+    Config.StableTokenV2 stableTokenConfig;
+  }
+
+  /**
+   * @dev Returns the populated configuration object for the PSO governance proposal.
+   */
+  function get(Contracts.Cache storage contracts) internal view returns (PSO memory config) {
+    config.poolConfig = PSOcUSD_PoolConfig(contracts);
+    config.rateFeedConfig = PHPUSD_RateFeedConfig();
+    config.stableTokenConfig = stableTokenKESConfig();
```
bayological, line 34:
```suggestion
    config.stableTokenConfig = stableTokenPSOConfig();
```

bayological (approved):

PR #211, opened 2024-08-19
celoscan contract verification check script

chapati23:
### Done
- [x] Wrote a simple `./bin/check-celoscan-verification.sh` script that accepts a celo mainnet contract address and will return 0 if the address is verified on celoscan and exit with 1 if it's not
- [x] Wrote a `./bin/pre-commit-check.sh` script that uses the celoscan verification script and runs it on all newly deployed contract deployments contained in a commit
   - It figures out newly deployed contracts by filtering the broadcast files for `transactionType: CREATE`

This will run for every broadcast file in a `git commit` to make sure we can't commit broadcast files of unverified contracts

### How to test
- Copy broadcast file with a contract deployment in it and `git add` it
- Run `./.husky/pre-commit`
- Check should run and pass
- Change the contractAddress in your staged broadcast file to some random unverified address
- Run `./.husky/pre-commit` and see that the check fails


bin/pre-commit-check.sh
```diff
@@ -0,0 +1,126 @@
...
+            ;;
+    esac
+}
+
+# Fetch addresses of newly deployed contracts from broadcast file
+fetch_addresses() {
+    cat "${broadcast_file}" |
+    jq -r '.transactions[] | select(.transactionType == "CREATE") | .contractAddress' |
+    grep -v null
+}
```
bowd, line 42: There's also factory contract case, where calling a function on a contract results in contract creation. 
Foundry verifies those contracts on Sourcify when executing the transaction, and our Sourcify->CeloScan script also picks them up. You can look at the `broadcasts/ExecuteProposal.sol/42220/MUGOV.json` to see the structure. Should be easy to concatenate.

bowd (approved):
Love it. I would potentially make it pre-push not pre-commit just to reduce friction, but looks great 👍

bin/pre-commit-check.sh
chapati23: good catch, i extended the jq expression to include factory deployments as well

philbow61 (approved):
Nice One 😎

PR #213, opened 2024-09-03
feat: convert uint ts 2 str

baroooo:
### Description

We were not properly encoding the TS to string before for testnet proposals, which was causing buggy proposal names for testnet proposals.

image

This pr fixes that issue by using the uint2str function.

### Other changes

no

### Tested

image

### Related issues

- Not tracked


denviljclarke (approved):
Beautiful 🤩

philbow61 (approved):
😎

PR #216, opened 2024-09-18
feat: add minipay airdrop csv generation script

### LINKED ISSUE #215: Script to generate the Minipay Airdrop allocation CSV
> [Notion Project Page](https://www.notion.so/mentolabs/MiniPay-Airdrop-Claiming-2d226ec13270425692ce4096dd302184?pvs=4)

### Overview

We have a [query in Dune](https://dune.com/queries/3932204?category=materialized_views&id=dune.mento-labs-eng.result_minipay_airgrab_start_balance) that gives us the building blocks for the airdrop allocation, specifically the `amount_transferred` and `avg_amount_held` for each address. We can use this to calculate:
- `mento allocation` = `amount_transferred` * 0.1 + `avg_amount_held`
- `cUSD allocation` = `mento allocation` * 0.1

The script should use the Dune API to query the data in chunks because there's a limit on how many datapoints can be served in one API call. In our case last I checked we could do ~30k rows per request (~120 requests needed). Then it would combine the data and output two csvs of (address,amount) for `mento` and `cUSD` that can be used as input to build out the merkle tree.

> Hint: when testing out the script, limit the number of rows you're fettching because each requests costs us money.

###  Acceptance criteria

- [ ] Script is written and pulls data from dune and saves it in the data dir
- [ ] Our other scripts are able to build a merkle tree from the output

philbow61:
### Description

This pr adds the script for generating the minipay airdrop csv from the dune query result.

### Tested

Ran the script and the size of csv entries matched the size of the query rows. Verified calculation is precise.
How to Test since running on the actual query could get expensive:
1. fork the dune query and limit the result to e.g 10( add `LIMIT 10` to the bottom of the query)
2. change the queryid in the script to match your fork of the query (Line 37 `const queryId = .... ;)`
3. change the limit in line 45 of the script `const limit = 2;`
(results in 2 rows being requested from the API per request)
5. verifying the results + number of entries in the csv 

### Related issues

- Fixes #215


RyRy79261 (approved):
LGTM, need to run test locally still

script/upgrades/MINIDROP/data/generate-airdrop-csv.ts
```diff
@@ -0,0 +1,97 @@
...
+  let rowsFetched = 0;
+  while (hasMorePages) {
+    let jsonResponse;
+    try {
+      const response = await fetch(uri, options);
+      jsonResponse = await response.json();
+      rowsFetched += jsonResponse.result.rows.length;
+      console.log(`⏳ ${rowsFetched}/${jsonResponse.result.metadata.total_row_count} records fetched`);
+    } catch (err) {
+      console.error(err);
```
bayological, line 58: If the API call fails this block will continue the execution, is that a problem here?

bayological (approved):
LGTM. Just one question

script/upgrades/MINIDROP/data/generate-airdrop-csv.ts
philbow61: I would say you are right a throw would be nicer here 👍

PR #219, opened 2024-09-24
Feat/update csv generation script to generate 2 csv

philbow61:
### Description

This Pr makes two changes to the CSV generation script:
- filtering out entries with less than 1 Mento earned 
- writing the result into two individual csv files one for mento and one for cUSD

### Other changes

- did some refactoring that hopefully improves code quality 🙃

### Tested

ran script testing this hasn't changed:
1. switch out query id with: 4096793
2. change limit from 10_000 to 10 
3. run script 

### Related issues

- Fixes https://github.com/mento-protocol/mento-core/issues/519

### DIFF
--- a/script/upgrades/MINIDROP/data/generate-airdrop-csv.ts
+++ b/script/upgrades/MINIDROP/data/generate-airdrop-csv.ts
@@ -20,6 +20,8 @@ interface Airdrop {
 config();
 
 const MENTO_THRESHOLD = new Decimal(1e18);
+const ROW_LIMIT = 10000;
+const QUERY_ID = 3932204; // https://dune.com/queries/3932204
 
 async function run() {
   const duneApiKey = process.env.DUNE_API_KEY;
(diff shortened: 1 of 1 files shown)


script/upgrades/MINIDROP/data/generate-airdrop-csv.ts
```diff
@@ -19,79 +19,106 @@ interface Airdrop {
...
+    path: cUSDFilepath,
+    header: [
+      { id: "address", title: "Address" },
       { id: "cUSDAllocation", title: "cUSD Allocation" },
     ],
-    append: false,
+    append: true,
   });
 
   const queryId = 3932204; // https://dune.com/queries/3932204
```
chapati23, line 54: nit: would pull this magic hardcoded const up next to the MENTO_THRESHOLD constant on top

script/upgrades/MINIDROP/data/generate-airdrop-csv.ts
```diff
@@ -19,79 +19,106 @@ interface Airdrop {
...
-
-  if (!duneApiKey) {
-    console.error("🚨 Missing DUNE_API_KEY environment variable");
-    process.exit(1);
-  }
-
   const options = { method: "GET", headers: { "X-DUNE-API-KEY": duneApiKey } };
   const limit = 10000;
```
chapati23, line 56: nit: same here, would pull out all hardcoded config to the top to make it clear those are hardcoded constants

chapati23 (approved):
lgtm, just some nit comments.

performance-wise it might be better to use streaming (i.e. `mentoStream = fs.createWriteStream(mentoFilePath)` instead of loading all rows into memory first but given we're gonna run this script just a couple times, this is probably unnecessary optimization

RyRy79261 (approved):

PR #239, opened 2024-11-05
Remove baklava references

chapati23:
### Description

fixes [mento-general#546](https://github.com/mento-protocol/mento-general/issues/546)

I wasn't sure if we wanted to keep the broadcast files so I left them in for now, but in theory we could also remove those unless somebody feels strongly opposed.

I tested running a few scripts locally and everything worked fine.


baroooo (approved):

script/dev/dev-DeployMockChainlinkAggregator.sol
```diff
@@ -10,7 +10,7 @@ import { MockChainlinkAggregator } from "contracts/MockChainlinkAggregator.sol";
 
 /**
  * Usage: yarn script:dev -n alfajores -s DeployMockChainlinkAggregator -r "run(string)" PHPUSD
- * Used to deploy mock Chainlink Aggregators to Alfajores and Baklava to be used
```
chapati23, line 13: interesting, the vs code search failed me!

nvtaveras (approved):

PR #243, opened 2024-11-07
feat: add script to deploy locking proxy admin

### LINKED ISSUE #240: Governance script that deploys a new proxyAdmin for Locking contracts
#### Description

Mento Governance and Locking contracts are currently in a circular dependency, any problem in the Locking contracts would block the governance from fixing the issue since the governance needs Locking to function properly to execute the upgrade. With the Celo L2 transition approaching with breaking changes to the Locking contracts, we should transfer the upgradability rights to another account. This is going to be temporary and it will enable us upgrading Locking contracts without depending Locking to be in the perfect state.

Implement a governance script that deploys a new Proxy Admin that is owned by MentoLabs multisig and transfer proxy ownership of Locking contracts to this new Proxy Admin contract.

#### Acceptance Criteria
- [ ] Script deploys a new Proxy Admin that is owned by MentoLabs Multisig
- [ ] Script transfers Locking proxy ownership to the new Proxy Admin
- [ ] Post checks

bayological:
### Description

- Adds script to deploy a proxy admin for the locking contract along with a governance script + checks

### Other changes

- None

### Tested

- Executed deployment, cgp & checks on local fork
image

### Related issues

- Fixes #240

### DIFF
--- a/remappings.txt
+++ b/remappings.txt
@@ -10,3 +10,4 @@ mento-core-2.2.0/=lib/mento-core-2.2.0/contracts/
 mento-core-2.3.1/=lib/mento-core-2.3.1/contracts/
 mento-core-2.5.0/=lib/mento-core-2.5.0/contracts/
 merkle-distributor/=lib/merkle-distributor/contracts/
+mento-core-2.6.0-oz/=lib/mento-core-2.6.0/lib/openzeppelin-contracts-next
--- a/remappings.txt
+++ b/remappings.txt
@@ -11,3 +11,4 @@ mento-core-2.3.1/=lib/mento-core-2.3.1/contracts/
 mento-core-2.5.0/=lib/mento-core-2.5.0/contracts/
 merkle-distributor/=lib/merkle-distributor/contracts/
 mento-core-2.6.0-oz/=lib/mento-core-2.6.0/lib/openzeppelin-contracts-next
+mento-core-2.6.0-tp/=lib/mento-core-2.6.0/lib/openzeppelin-contracts-next/contracts/proxy/transparent
--- a/remappings.txt
+++ b/remappings.txt
@@ -11,4 +11,3 @@ mento-core-2.3.1/=lib/mento-core-2.3.1/contracts/
 mento-core-2.5.0/=lib/mento-core-2.5.0/contracts/
 merkle-distributor/=lib/merkle-distributor/contracts/
 mento-core-2.6.0-oz/=lib/mento-core-2.6.0/lib/openzeppelin-contracts-next
-mento-core-2.6.0-tp/=lib/mento-core-2.6.0/lib/openzeppelin-contracts-next/contracts/proxy/transparent
--- a/remappings.txt
+++ b/remappings.txt
@@ -11,3 +11,4 @@ mento-core-2.3.1/=lib/mento-core-2.3.1/contracts/
 mento-core-2.5.0/=lib/mento-core-2.5.0/contracts/
 merkle-distributor/=lib/merkle-distributor/contracts/
 mento-core-2.6.0-oz/=lib/mento-core-2.6.0/lib/openzeppelin-contracts-next
+mento-core-2.6.0/=lib/mento-core-2.6.0/contracts/
--- a/script/upgrades/MU09/MU09.sol
+++ b/script/upgrades/MU09/MU09.sol
@@ -1,22 +1,24 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 // solhint-disable func-name-mixedcase, contract-name-camelcase, function-max-lines, var-name-mixedcase
-pragma solidity ^0.5.13;
+pragma solidity ^0.8;
 pragma experimental ABIEncoderV2;
 
 import { GovernanceScript } from "script/utils/mento/Script.sol";
 import { console } from "forge-std/console.sol";
-import { Contracts } from "script/utils/Contracts.sol";
-import { Chain } from "script/utils/Chain.sol";  
+import { Contracts } from "script/utils/mento/Contracts.sol";
+import { Chain } from "script/utils/mento/Chain.sol";
 
 import { IGovernanceFactory } from "script/interfaces/IGovernanceFactory.sol";
 import { IMentoUpgrade, ICeloGovernance } from "script/interfaces/IMentoUpgrade.sol";
 
-interface IOwnableLite { 
+interface IOwnableLite {
   function transferOwnership(address recipient) external;
 }
 
 contract MU09 is IMentoUpgrade, GovernanceScript {
   using Contracts for Contracts.Cache;
+
+  bool public hasChecks = true;
 
   address public mentoGovernor;
   address public lockingProxyAdmin;
@@ -76,7 +78,7 @@ contract MU09 is IMentoUpgrade, GovernanceScript {
     _transactions[0] = ICeloGovernance.Transaction(
       0,
       lockingProxy,
-      abi.encodeWithSelector(IOwnableLite(0).transferOwnership.selector, lockingProxyAdmin)
+      abi.encodeWithSelector(IOwnableLite.transferOwnership.selector, lockingProxyAdmin)
     );
 
     return _transactions;
--- a/script/upgrades/MU09/MU09.sol
+++ b/script/upgrades/MU09/MU09.sol
@@ -46,11 +46,9 @@ contract MU09 is IMentoUpgrade, GovernanceScript {
 
     // Get newly deployed LockingProxyAdmin address
     lockingProxyAdmin = contracts.deployed("ProxyAdmin");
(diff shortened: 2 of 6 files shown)


script/upgrades/MU09/MU09.sol
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  function buildProposal() public returns (ICeloGovernance.Transaction[] memory) {
+    ICeloGovernance.Transaction[] memory _transactions = new ICeloGovernance.Transaction[](1);
+
+    // Create transaction to transfer proxy ownership
+    _transactions[0] = ICeloGovernance.Transaction(
+      0,
+      lockingProxy,
+      abi.encodeWithSelector(IOwnableLite.transferOwnership.selector, lockingProxyAdmin)
```
baroooo, line 81: What we do here is to transfer the contract ownership and this does not give the upgradability rights.
We should instead call changeProxyAdmin on the old proxyAdmin, locking proxy being the first param.

script/upgrades/MU09/MU09Checks.sol
```diff
@@ -0,0 +1,70 @@
...
+
+    address lockingProxyAdminOwner = IOwnableLite(lockingProxyAdmin).owner();
+    require(lockingProxyAdminOwner == mentoLabsMultisig, "LockingProxyAdmin owner is not MentoLabsMultisig");
+    console.log(unicode"🟢 LockingProxyAdmin owner is MentoLabsMultisig: %s", lockingProxyAdminOwner);
+  }
+
+  function verifyLockingProxyOwnership() public {
+    console.log("\n== Verifying locking proxy ownership: ==");
+
+    address lockingProxyOwner = IOwnableLite(lockingProxy).owner();
```
baroooo, line 66: This check should be replaced by a getProxyAdmin call to the newProxy admin, and returned value should be the new proxy admin address.
The same call to the old proxy admin should be reverted.

baroooo (changes requested):

script/upgrades/MU09/MU09.sol
```diff
@@ -0,0 +1,101 @@
...
+    // Load the deployed ProxyAdmin contract
+    contracts.loadSilent("MU09-Deploy-LockingProxyAdmin", "latest");
+  }
+
+  function prepare() public {
+    loadDeployedContracts();
+
+    // Get newly deployed LockingProxyAdmin address
+    lockingProxyAdmin = contracts.deployed("ProxyAdmin");
+    require(lockingProxyAdmin != address(0), "LockingProxyAdmin address not found");
```
philbow61, line 49: nit: deployed already verifies address not being address(0)

script/upgrades/MU09/MU09Checks.sol
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  function verifyLockingProxyAdminIsNewLockingProxyAdmin() public {
+    console.log("\n== Verifying locking proxy admin is new locking proxy admin: ==");
+
+    address lockingProxyAdmin = IProxyAdminLite(newLockingProxyAdmin).getProxyAdmin(lockingProxy);
+    require(lockingProxyAdmin == newLockingProxyAdmin, "LockingProxyAdmin is not the new LockingProxyAdmin");
+    console.log(unicode"🟢 LockingProxyAdmin is new LockingProxyAdmin: %s", lockingProxyAdmin);
+  }
+}
```
philbow61, line 72: potentially we could add another check here ensuring that the mentolabsMultisig can upgrade the the locking implementation 👍

philbow61 (approved):

script/upgrades/MU09/MU09.sol
```diff
@@ -0,0 +1,101 @@
...
+interface IProxyAdminLite {
+  function getProxyAdmin(address proxy) external view returns (address);
+
+  function changeProxyAdmin(address proxy, address newAdmin) external;
+}
+
+contract MU09 is IMentoUpgrade, GovernanceScript {
+  using Contracts for Contracts.Cache;
+
+  bool public hasChecks = true;
```
Andrew718PLTS, line 23: Do we need this field in public? if we do, where are gonna use it?

script/upgrades/MU09/MU09.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    vm.startBroadcast(Chain.deployerPrivateKey());
+    {
+      // TODO: Change this to the forum post URL
+      createProposal(_transactions, "https://CHANGE-ME-PLEASE", mentoGovernor);
+    }
+    vm.stopBroadcast();
+  }
+
+  function buildProposal() public returns (ICeloGovernance.Transaction[] memory) {
```
Andrew718PLTS, line 82: It is a public function, do we use it somewhere besides the run function?

script/upgrades/MU09/MU09.sol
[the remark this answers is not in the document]
bayological: Used by the tooling to run the cgp checks when running the simulation. SimulateUpgrade.sol ln 25

script/upgrades/MU09/MU09.sol
bayological: This is intentional. This function is also called by the tooling.

remappings.txt
```diff
@@ -9,4 +9,6 @@ mento-core-2.1.0/=lib/mento-core-2.1.0/contracts/
 mento-core-2.2.0/=lib/mento-core-2.2.0/contracts/
 mento-core-2.3.1/=lib/mento-core-2.3.1/contracts/
 mento-core-2.5.0/=lib/mento-core-2.5.0/contracts/
-merkle-distributor/=lib/merkle-distributor/contracts/
\ No newline at end of file
+merkle-distributor/=lib/merkle-distributor/contracts/
+mento-core-2.6.0-oz/=lib/mento-core-2.6.0/lib/openzeppelin-contracts-next
+mento-core-2.6.0-tp/=lib/mento-core-2.6.0/lib/openzeppelin-contracts-next/contracts/proxy/transparent
```
baroooo, line 14: nit: I am not sure about creating remappings for sub dirs like this one

script/upgrades/MU09/MU09Checks.sol
```diff
@@ -0,0 +1,86 @@
...
+pragma experimental ABIEncoderV2;
+
+import { GovernanceScript } from "script/utils/mento/Script.sol";
+import { console } from "forge-std/console.sol";
+import { Contracts } from "script/utils/mento/Contracts.sol";
+import { Test } from "forge-std/Test.sol";
+
+import { IGovernanceFactory } from "script/interfaces/IGovernanceFactory.sol";
+import { ITransparentUpgradeableProxy } from "mento-core-2.6.0-tp/TransparentUpgradeableProxy.sol";
+import { ProxyAdmin } from "mento-core-2.6.0-tp/ProxyAdmin.sol";
```
baroooo, line 12: nit: we use ProxyAdmin in checks but in MU09 we use IProxyAdminLite

script/upgrades/MU09/MU09Checks.sol
```diff
@@ -0,0 +1,86 @@
...
+    console.log("\n== Verifying the multisig can successfuly upgrade implementation: ==");
+
+    address fakeImplementation = mentoLabsMultisig;
+    vm.startPrank(mentoLabsMultisig);
+
+    // Verify that the upgrade event is emitted with the fake implementation
+    vm.expectEmit(true, true, true, true);
+    emit Upgraded(fakeImplementation);
+
+    ProxyAdmin(newLockingProxyAdmin).upgrade(ITransparentUpgradeableProxy(lockingProxy), fakeImplementation);
```
baroooo, line 82: MentoLabsMultisig configured for Alfajores is not a contract so this should fail: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/ERC1967/ERC1967Utils.sol#L54

remappings.txt
bayological: Agreed. I did this because the import path was too long without it. The alternative is changing the max line length in the linter + ci

script/upgrades/MU09/MU09Checks.sol
bayological: Sure, I ran this on a mainnet fork so it did not fail. I can use a contract that exists on both if you think that's necessary?

script/upgrades/MU09/MU09Checks.sol
bayological: Won't change unless you think it's necessary. I switched to using the full implementation in the checks when adding the upgrade check

script/upgrades/MU09/MU09Checks.sol
baroooo: yeah lets do it so it runs checks successfully after executing the proposal on alfajores

baroooo (approved):

PR #244, opened 2024-11-13
feat: add celo transfer to MU08

philbow61:
### Description

This updates MU08 to add the CELO return from the Mento Reserve to Celo Governance (https://celo.stake.id/#/proposal/102). There are a lot of transactions in order to accomplish this, but the gist of it is:

- A new Reserve proxy ("Custody Reserve") has been deployed ([alfajores](https://alfajores.celoscan.io/address/0x43285A2FF682589d7c7260CD2Ab3dd9456Eaa204), [mainnet](https://celoscan.io/address/0x1E0433C1769271ECcF4CFF9FDdD515eefE6CdF92)), with the following configuration:
  - Celo Governance as both a spender and an "other reserve" address (meaning governance can initiate tx's to send any CELO in the reserve to itself)
  - CELO as a collateral asset, with a 100% spending ratio
  - Mento Governance as the owner of the proxy contract, as an additional safe-guard mechanism that can intervene if needed
- 80M CELO is transferred to the new Custody Reserve
- 20M initial CELO is transferred to Celo Governance as part of the separation CGP. The rest will sit in the new reserve for Celo Governance to use in the future as they seem fit.

### Other changes

N/A

### Tested

- Checks are written to ensure the transfers/wiring happens correctly, including checks for the operations that remove some of the temporary configuration that was added in order to accomplish the transfer (e.g. remove celo gov as a spender on the Mento Reserve, remove Custody reserve as an "other address" from Mento Reserve).
- Simulation is green on Alfajores

### Related issues

- Fixes https://github.com/mento-protocol/mento-general/issues/589


script/upgrades/MU08/MU08.md
```diff
@@ -1,9 +1,12 @@
 ### MU08
 
-Transfer ownership of the MentoProtocol contracts to Mento Governance.
+Transfer ownership of the MentoProtocol contracts to Mento Governance. Return of 80 mio CELO to Celo Governance.
```
baroooo, line 3: is `mio` a correct usage? I would prefer `M` or `mil`

baroooo (approved):
✅

PR #245, opened 2024-11-25
fix: MU08 CELO return amounts

nvtaveras:
### Description

This updates the amount of CELO that will be return in celo governance as part of MU08. We went with a rough number of ~80M CELO but the exact amount is `82,406,987`.

### Other changes

N/A

### Tested

Simulation is green on Alfajores


philbow61 (approved):
LGTM please also update the readme of MU08

bayological (approved):

PR #250, opened 2024-12-17
feat: MGP03 and MGP04

baroooo:
### Description

This adds MGP03 (Set the Mento Labs Multisig ahead of L2 transition) and MGP04 (Increase the Governance voting period for the L2 transition), both which I already executed on Alfajores.

MGP04 can't be simulated, because the [setVotingPeriod function](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/governance/extensions/GovernorSettingsUpgradeable.sol#L87-L89) inside Governor is protected by [a special modifier](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/governance/GovernorUpgradeable.sol#L73-L86) that ensures the call can only be made from within a proposal, so when we try to [simulate it](https://github.com/mento-protocol/mento-deployment/blob/main/script/utils/mento/GovernanceHelper.sol#L54-L58) by pranking as the timeLock and calling the governor contract directly, the call reverts. The way to do it would be to change the simulation code to submit the transaction as a proposal, vote, and execute it. IMHO the simulation is not a big deal for this proposal as it's a single call, and the only important thing to check is the value of the new voting period.

### Other changes

I updated the `MentoLabsMultisig` address on Alfajores to the deployer account, as we had it set to a different EOA that we created for testing the token deployment (because we wanted to distribute MENTO tokens to different addresses). We have control of that other EOA as well, but using the deployer account feels more correct.

### Tested

Executed both on Alfajores

MGP03
https://alfajores.celoscan.io/tx/0x4f40b7bb83ed5459b97626e6ede346546fd83781a911aecb98faee1161c80a9f
> ❯ cast call 0x537CaE97C588C6DA64A385817F3D3563DDCf0591 "mentoLabsMultisig()(address)" --rpc-url $ALFAJORES_RPC_URL
0x56fD3F2bEE130e9867942D0F463a16fBE49B8d81

MGP04
https://alfajores.celoscan.io/tx/0x129e7f6c740a26c2e5d976ae64809b82089de9c561ad67a2fb10bd935a7c37dd
> ❯ cast call 0x558e92236f85Bb4e8A63ec0D5Bf9d34087Eab744 "votingPeriod()(uint256)" --rpc-url $ALFAJORES_RPC_URL
300


script/upgrades/MGP04/MGP04.sol
```diff
@@ -0,0 +1,64 @@
...
+      // TODO: Change this to the forum post URL
+      createProposal(_transactions, "updateWithUrl", mentoGovernor);
+    }
+    vm.stopBroadcast();
+  }
+
+  function buildProposal() public returns (ICeloGovernance.Transaction[] memory) {
+    ICeloGovernance.Transaction[] memory _transactions = new ICeloGovernance.Transaction[](1);
+
+    (uint256 currentVotingPeriod, uint256 newVotingPeriod) = Chain.isCelo() ? (120960, 604800) : (60, 300);
```
nvtaveras, line 52: @baroooo Was debating on whether to set this to 8 days already because of the governance transition. WDYT?

nvtaveras: discussed offline, will leave as is until MU08 executes. Added some TODO's to check this later on

script/upgrades/MGP03/MGP03Checks.sol
```diff
@@ -0,0 +1,75 @@
...
+  function l2TransitionBlock() external view returns (uint256);
+
+  function paused() external view returns (bool);
+}
+
+contract MGP03Checks is GovernanceScript, Test {
+  using Contracts for Contracts.Cache;
+
+  address public mentoLabsMultisig;
+  address public mentoGovernor;
```
baroooo, line 29: nit: this is not used

baroooo (commented):
looks good but I can not approve

script/upgrades/MGP03/MGP03Checks.sol
nvtaveras: removed. also, will approve the PR on your behalf

nvtaveras (approved):
👍🏽 from @baroooo as he was the one who created the PR

philbow61 (approved):
LGTM 
but let's remove the MU04 checks since they can't be run

PR #254, opened 2025-02-03
Feat: Locking V2 implementation deployment script 

RyRy79261:
### Description

This PR adds a script to allow for the deployment of the new LockingV2 contracts.

To verify this all works as expected, I ran against a local Alfajores fork. 

The process here, includes running the init function on the contract, while this has no final effect on the usage of the new contract by the proxy, it was identified as a function that could be called by the Hats audit. 

Below is the script output 

image

Here is a owner of call to verify the init was called successfully 
image


script/dev/dev-UpgradeLockingContract.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
nvtaveras, line 1: We shouldn't copy over these contracts and should import them from mento-core instead. e.g.: https://github.com/mento-protocol/mento-deployment/blob/0efbfe2df2844e71570608cb0478a771f657c881/script/upgrades/MU08/deploy/MU08-00-Create-Proxies.sol#L8

We can create a new release in mento-core and tag it once the audit changes are merged into develop

remappings.txt
```diff
@@ -12,3 +12,33 @@ mento-core-2.5.0/=lib/mento-core-2.5.0/contracts/
 merkle-distributor/=lib/merkle-distributor/contracts/
 mento-core-2.6.0-oz/=lib/mento-core-2.6.0/lib/openzeppelin-contracts-next
 mento-core-2.6.0/=lib/mento-core-2.6.0/contracts/
+openzeppelin-solidity/=lib/mento-core-2.0.0/lib/openzeppelin-contracts/
```
nvtaveras, line 15: some of these seem to be duplicated and not very relevant (i.e. `chainlink-brownie-contracts`, foundry-chainlink-toolkit`) maybe remove this file from the changes or just leave the ones that are needed?

script/dev/dev-UpgradeLockingContract.sol
```diff
@@ -0,0 +1,37 @@
...
+    contracts.load("MUGOV-00-Create-Factory", "latest");
+    address lockingV2;
+    address mentoToken = IGovernanceFactory(contracts.deployed("GovernanceFactory")).mentoToken();
+    vm.startBroadcast(ChainLib.deployerPrivateKey());
+    {
+      lockingV2 = address(new Locking());
+      console2.log("----------");
+      console2.log("LockingV2 deployed at: ", lockingV2);
+      console2.log("----------");
+      Locking(lockingV2).__Locking_init(IERC20Upgradeable(mentoToken), 212, 0, 1);
```
nvtaveras, line 31: lets maybe move these parameters into variables and add a comment for context to specify they are taken from mainnet?

RyRy79261: Roger that

remappings.txt
RyRy79261: @nvtaveras So essentially I ran the `forge remappings >> remappings.txt` for this.

I'd assume that if an entry is added here that there is a potential for it to be needed later with little affect on performance. 

That being said I am unsure of how likely such a situation is

script/dev/dev-UpgradeLockingContract.sol
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity 0.8.18;
+
+import { console2 } from "forge-std/Script.sol";
+import { Script } from "script/utils/mento/Script.sol";
+import { Chain as ChainLib } from "script/utils/mento/Chain.sol";
+import { Contracts } from "script/utils/mento/Contracts.sol";
+import { IRegistry } from "script/interfaces/IRegistry.sol";
```
nvtaveras, line 8: I think this is unused?

nvtaveras (approved):

script/dev/dev-UpgradeLockingContract.sol
```diff
@@ -0,0 +1,30 @@
...
+import { Chain as ChainLib } from "script/utils/mento/Chain.sol";
+import { Contracts } from "script/utils/mento/Contracts.sol";
+import { Locking } from "mento-core-2.6.2/governance/locking/Locking.sol";
+
+/**
+ * Usage: yarn script:dev -n alfajores -s UpgradeLockingContract -r "run()"
+ * Used to deploy the Locking V2 implementation
+ * ===========================================================
+ */
+contract UpgradeLockingContract is Script {
```
baroooo, line 15: deploy new locking implementation would be a better name for this script

script/dev/dev-UpgradeLockingContract.sol
```diff
@@ -0,0 +1,30 @@
...
+  using Contracts for Contracts.Cache;
+
+  function run() public {
+    contracts.load("MUGOV-00-Create-Factory", "latest");
```
baroooo, line 19: factory is not used

baroooo (approved):
looks good apart from some nits

PR #255, opened 2025-02-05
feat: cGHS launch deployment scripts

bayological:
### Description

This is the current state we're dealing with at the time of this PR creation:
- Mento contracts on Alfajores have had their ownership transferred to Mento governance
- Mento contracts on Mainnet are still under Celo governance
- **FeeCurrencyWhitelist** is no longer being used on **Alfajores** and has been replaced with the [FeeCurrencyDirectory](https://github.com/celo-org/celo-monorepo/blob/release/core-contracts/12/packages/protocol/contracts-0.8/common/FeeCurrencyDirectory.sol)
- **FeeCurrencyDirectory** does not yet exist on **mainnet** so the **FeeCurrencyWhitelist** contract is still being used on mainnet

Because of the above, this proposal is broken into two scripts(cGHSMento & cGHSCelo) for the two governance systems we are now dealing with.

Each of the scripts also includes checks for the network, to determine which transactions should be included based on the network. (e.g. on alfajores we use the FeeCurrencyDirectory but on Mainnet we use the FeeCurrencyWhitelist)

PR adds the following:

- Scripts for the creation & configuration of pools to support the launch of cGHS
- Post deployment verification scripts

### Other changes

- None

### Tested

- Alfajores - Celo Governance Proposal Simulation
image

- Mainnet - Celo Governance Proposal Simulation
- No rates so can't simulate at this stage

### Backwards compatibility

N/A


philbow61 (commented):
we need to set the referenceRateResetFrequency to 6 mins to match the config in SortedOracles

philbow61 (approved):

baroooo (approved):

PR #257, opened 2025-02-07
fix: update MU08 CELO return amount

nvtaveras:
### Description

The previous CELO amount to be returned to the Celo community (`82_406_987`) didn't include the CELO that was sent to the Reserve via the Reserve bolstering mechanism, which was an additional 3.5M CELO. This updates the amount to 85.9M in total.

The additional CELO was taken from on-chain epoch rewards data, for a total sum of `3,534,512.340972869827370586` CELO.

### Other changes

N/A

### Tested

Simulation green on Mainnet.


bowd (approved):

philbow61 (approved):
LGTM

PR #268, opened 2025-03-28
feat: add scripts to launch CHF, NGN, JPY

bayological:
### Description

This adds the script to generate the proposal for launching three new stables. 

### Tested

TODO - Depends on the deployment script s to be created as part of FX02.

#### Checking the limits

- For the recent stable token launches, all pools use the same limits ($200k, $1M, and $5M)
- The actual value for the limits are denominated in the asset using the USD to asset rate
- For limits where the rate is a decimal (e.g. USD to CHF == 0.88) I've scaled down the dollar limit and scaled up the rate. So instead of 0.88 * 200_000 we use 88 * 2_000
- (0.88 * 200,000) == (88 * 2,000) 

To get the rates you can use xe.com.


philbow61 (approved):
👍

nvtaveras (approved):

PR #269, opened 2025-03-31
feat: add mgp06

baroooo:
### Description

This proposal removes mentolabs multisig address from locking contract.

Details: script/upgrades/MGP06/MGP06.md

### Tested

Tested on a simulation and on alfajores


broadcast/MGP06.sol/44787/run-latest.json
nvtaveras, line 1: lets delete this as we don't tend to add broadcast files for proposals/executions

nvtaveras (approved):

foundry.toml
```diff
@@ -16,6 +16,7 @@ fs_permissions = [
     { access = "read", path = "./data" },
     { access = "read", path = "./script/upgrades/dependencies.json" },
     { access = "read", path = "./script/upgrades/MINIDROP/data/" },
+    { access = "read", path = "./script/upgrades/MGP06/" },
```
bayological, line 19: Should we instead grant read access to the upgrades directory here?

bayological (approved):

